PR #54, opened 2024-01-09
feat: Allow limited functions to users who have server scripts disabled

surajshetty3416:
Allow limited functions to users who have server scripts disabled

```
frappe.db.count
frappe.db.exists
frappe.db.get_all (alias of safe_get_all)
frappe.db.get_list (alias of safe_get_all)
frappe.make_get_request (alias of make_safe_get_request)
frappe.get_doc (alias of get_doc_as_dict)
frappe.get_cached_doc (alias of get_cached_doc_as_dict)
```

### DIFF
--- a/builder/utils.py
+++ b/builder/utils.py
@@ -8,7 +8,8 @@ from frappe.utils.safe_exec import (
 	NamespaceDict,
 	get_python_builtins,
 	safe_exec_flags,
-	get_safe_globals
+	get_safe_globals,
+	SERVER_SCRIPT_FILE_PREFIX
 )
 
 from RestrictedPython import compile_restricted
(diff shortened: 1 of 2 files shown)


builder/utils.py
```diff
@@ -0,0 +1,115 @@
...
+):
+	exec_globals = get_safer_globals()
+	if _globals:
+		exec_globals.update(_globals)
+
+	with safe_exec_flags():
+		# execute script compiled by RestrictedPython
+		exec(
+			compile_restricted(
+				script, filename="<data-script>", policy=FrappeTransformer
```
ankush, line 109: Add a name here so it's easy to read. 

Also use `serverscript` somewhere in filename. it's used by QB to check if you're in server script or not.

ankush: https://github.com/ankush/frappe/blob/a6e2c38635e3dbc3a92fc7b5ad0b6013aecbf15c/frappe/query_builder/utils.py#L127-L128

surajshetty3416: https://github.com/frappe/builder/pull/54/commits/bdf9213ed2db52a744b9cfa6039d0415c7aad7bf

PR #79, opened 2024-02-21
feat: Page templates

surajshetty3416:
- Create new page from predefined template


surajshetty3416:
Merging this since this has usable block templates & refactored code. Page templates will be disabled/inactive for now till we have good templates.

federicocalvo:
@surajshetty3416 This PR was the most important pr as far as the masses were concerned. Because it allowed you to start from a base, see good examples, etc., it is still a very difficult tool for end users, even for advanced users.

While the blocks obviously solved or come to solve, I think this was at another level, and would greatly slow down the adoption.

Hopefully they will resume this, the example that was very good, I imagine that having 3 of these, it would be crazy to explore them, use creativity, etc..

PR #109, opened 2024-03-12
refactor: enable change tracking on Builder Pages

batonac:
In the future, it would be great to have a UI for versioning management, but for now, I think we should start tracking changes, so there's some kind of fallback when things to awry.

### DIFF
--- a/builder/builder/doctype/builder_page/builder_page.json
+++ b/builder/builder/doctype/builder_page/builder_page.json
@@ -170,5 +170,6 @@
  "sort_field": "modified",
  "sort_order": "DESC",
  "states": [],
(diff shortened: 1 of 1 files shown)


surajshetty3416:
Hi @batonac,

Every minor change is auto saved in builder (except for component) hence having versions for all of them will be point less. I'll suggest not enable this right now as it'll take lot of space in the db which is mostly unnecessary.

I do have a plan to introduce some kind of versioning (kind of similar to google docs... a good chunk of changes can be part of single version.

batonac:
Makes sense. Thanks for the clarification!

Issue #96, opened 2024-03-05
Bug: multiline text element is translated as two separate elements


the3dfox:
Awesome new text style feature including h1, h2... 

I see a problem though when having a line break in a text element:

![image](https://github.com/frappe/builder/assets/84663901/44d81745-7c35-4578-883f-802ee2bab956)

this translates to two separate elements though:
![image](https://github.com/frappe/builder/assets/84663901/4e132e4c-1bef-4b48-857a-f7db5d7c2685)

Which is especially problematic when the text is a heading, because the result is acutally two headings:
![image](https://github.com/frappe/builder/assets/84663901/fb15eb9d-5cc6-4797-8530-54bd098f4d5f)

valid implementations seem to be:

`<h1>Line 1 <br/> Line 2</h1>`
`<h1>Line 1<br>Line 2</h1>`
`<h1>Line 1 <span style = "display: block;">Line 2</span></h1>`

results can be checked here:
https://www.seoreviewtools.com/html-headings-checker/

PR #142, opened 2024-05-05
fix: Update tag on toggling heading

surajshetty3416:
Previously, we used to wrap content inside new heading tags and update innerHTML (using tiptap's utility function) instead of updating the main block's element. This fix will update the tag of the main text block.

**Before:** 
Screenshot 2024-05-05 at 4 03 43 PM

**After:**
Screenshot 2024-05-05 at 4 02 16 PM

**Note:** To use this fix, you'll have to recreate your heading block.

closes: https://github.com/frappe/builder/issues/96


surajshetty3416:
@mergifyio backport master

PR #164, opened 2024-05-29
fix: Allow dynamic styles using Data Key

surajshetty3416:
This feature allows us to fetch dynamic value for a CSS property

https://github.com/frappe/builder/assets/13928957/30c9f59c-2132-4457-b5b7-154e7c4e64cb


surajshetty3416:
@mergifyio backport master

PR #170, opened 2024-06-10
fix: Add rel="nofollow" attribute to indicate not to follow that link

surajshetty3416:
To prevent search engines from crawling editor link.


surajshetty3416:
@mergifyio backport master

PR #200, opened 2024-08-16
fix: consider configured builder app path in builder page link

ruchamahabal:
The "Open in Builder" link does not honor configured builder path from site config

https://github.com/user-attachments/assets/8b740a3f-3dc5-4826-aeea-a0dde0c22252


surajshetty3416:
Fixed with https://github.com/frappe/builder/commit/a4a576128b01a2df942dddeba971039c84431229. It did not require additional request. 

Since the approach was different, committed it directly instead of pushing here.

PR #203, opened 2024-08-23
fix: generate the route value only if self.route is empty

omkardeekonda:
Creating this PR to allow route auto-generation only if the route is not already set.

To maintain all the builder pages in the repository we used the `bench export-fixtures` option.
Upon using the `bench migrate` on production environment all the routes auto-generated. Routes from the fixtures are not picked up.
To resolve the issue added a check in the `before_insert` method of `builder_page.py`.
This is to ensure that the fixtures import doesn't mess up the routes.


surajshetty3416 (approved):

omkardeekonda:
Thanks for merging @surajshetty3416 :+1:

PR #223, opened 2024-10-19
fix: Misc

surajshetty3416:
- Avoid text selection while container resize or color selection
- Show current color inside color selector
- Fixed the conversion of selected text to link feature


surajshetty3416:
@Mergifyio backport master

PR #304, opened 2025-01-27
feat: Show drop target placeholder during drag & drop

ruchamahabal:
### Before:

Hard to figure out where my component/block will drop

https://github.com/user-attachments/assets/a6491ec2-220f-4d44-830a-86377288a3d8

### After:

Porting from studio: https://github.com/frappe/studio/pull/14 & https://github.com/frappe/studio/pull/15

Shows a placeholder for drag target

https://github.com/user-attachments/assets/209b16a4-0ff3-4bf8-b4db-ea47fee22b23

Flips placeholder orientation as per layout direction — column/row to avoid too much layout shifting and follow general flow of elements in the container. Eg: This is a Quick Stack with flex direction = row, so elements will be added vertically, so placeholder flips accordingly

https://github.com/user-attachments/assets/9822e2c7-b27a-465c-9efc-05292fbf8967

### How

**dragstart**

- Inserts a placeholder div in DOM on dragstart (dashed border with some margin to differentiate it from the selected block borders)

**dragover**
- Finds the parent block (that can have children) based on mouse position
- Finds the exact index at which to drop the component/block
	- Finds midpoints of all child elements as per layout direction
	- Finds the closest child to the mouse position
	- If the mouse is to the left/top of the midpoint, inserts placeholder before that child
	- If the mouse is to the right/below the midpoint, inserts placeholder after that child
	- Index defaults to end of the parent container if no children are found
- updates drag state in store

**drop**
- fetches state from store, inserts the element at that position

**dragend**
- resets drag state
- removes placeholder from DOM

<hr>

- [ ] Handle/refactor replace block case
- [ ] Handle sensitivity while dropping on shared edges between 2 blocks 
- [ ] Test grid layouts

### DIFF
--- a/frontend/src/components/BlockEditor.vue
+++ b/frontend/src/components/BlockEditor.vue
@@ -52,6 +52,7 @@ const showResizer = computed(() => {
 	return (
 		!props.block.isRoot() &&
 		!props.editable &&
+		!store.isDragging &&
 		isBlockSelected.value &&
 		!blockController.multipleBlocksSelected() &&
 		!props.block.getParentBlock()?.isGrid() &&
@@ -93,6 +94,7 @@ const showPaddingHandler = computed(() => {
 	return (
 		isBlockSelected.value &&
 		!resizing.value &&
+		!store.isDragging &&
 		!props.editable &&
 		!blockController.multipleBlocksSelected() &&
 		!props.block.isSVG() &&
@@ -104,6 +106,7 @@ const showMarginHandler = computed(() => {
 	return (
 		isBlockSelected.value &&
 		!props.block.isRoot() &&
+		!store.isDragging &&
 		!resizing.value &&
 		!props.editable &&
 		!blockController.multipleBlocksSelected() &&
@@ -120,6 +123,7 @@ const showBorderRadiusHandler = computed(() => {
 		!props.block.isSVG() &&
 		!props.editable &&
 		!resizing.value &&
+		!store.isDragging &&
 		!blockController.multipleBlocksSelected()
 	);
 });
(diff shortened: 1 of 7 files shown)


surajshetty3416:
Awesome!

---

Few additional things to take care of
- [ ] Fix drop in other breakpoints
- [ ] Fix dragImage while dragging the edge of a template
	
	https://github.com/user-attachments/assets/263da1e1-2c54-4771-bae4-feeb475a9129
- [ ] this flicker... possible to fix?
	
	https://github.com/user-attachments/assets/527addaa-c4b9-456f-aa57-3972268c4a08

ruchamahabal:
> Fix drop in other breakpoints

https://github.com/user-attachments/assets/3833b5bb-9eba-4a57-926d-3aa10f2c99df

>  Fix dragImage while dragging the edge of a template

https://github.com/user-attachments/assets/29fa95ca-d429-4402-8228-bae3684a4fa0

>  this flicker... possible to fix?

Might not be fixed completely but drop target won't be recomputed if mouse isn't moving

https://github.com/user-attachments/assets/05bce8e1-4a24-4f06-bedf-77808da475dc

> Replace block

https://github.com/user-attachments/assets/538e71ab-2613-4a5e-91ae-709ee4530219

> File drop placeholder

https://github.com/user-attachments/assets/b897befa-224a-4b1e-8d8c-10e7bd419961

> Handle sensitivity while dropping on shared edges between 2 blocks
This & flickering is still not completely solved. Partially solved with skipping updates when mouse pointer does not move. Can be tackled later IG

surajshetty3416 (approved):

PR #325, opened 2025-03-17
fix: Hide edit icon for Website Users

bvisible:
This PR fixes an issue where the Builder edit icon was visible to users who don't have sufficient permissions to edit pages, such as Website Users. 

## Changes
- Added an additional check in the editor icon condition to hide it for Website Users
- The icon will now only be displayed for System Users who have the necessary privileges

## Why
Previously, the edit icon was being displayed for all logged-in users, including Website Users who don't have permission to use Builder. This was confusing as clicking the icon would lead them to an access denied page.

## Testing
- Login as a Website User and verify that the edit icon doesn't appear
- Login as a System User and verify that the edit icon still appears properly

### DIFF
--- a/builder/templates/generators/webpage_scripts.html
+++ b/builder/templates/generators/webpage_scripts.html
@@ -15,7 +15,7 @@
 {% endif %}
 {% if not preview %}
 <script>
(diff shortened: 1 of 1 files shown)


surajshetty3416:
@Mergifyio backport master

PR #423, opened 2025-10-20
feat: add block props and block js scripts

stravo1:
This PR introduces **block-level props** and **block-specific JavaScript scripts**.

Each block now has two new properties: **Block Script** and **Props**.

* **Block Script:**
  These are JavaScript code snippets associated with an individual block, unlike **Client Scripts** which apply to an entire page.
  Block Scripts make it easier to manage reactivity by eliminating the need for multiple Client Scripts and by localizing logic to the block itself.

  Within a Block Script, two special keywords are available:

  * `this`: Refers directly to the DOM node corresponding to the block.
    Example:

    ```js
    this.addEventListener('click', ...)
    ```

    attaches an event listener to the block itself.
  * `props`: Provides access to the properties passed to the block. Props are discussed below.

* **Props:**
  Props work similarly to those found in component-based JavaScript frameworks such as Vue or React. They serve as an interface between the “outside world” and the block script.

  There are three types of props:

  * **static** - Constant literal values.
  * **dynamic** - Derived from Data Scripts.
  * **inherited** - Passed down from parent blocks.

As mentioned earlier, props can be accessed within the Block Script using the `props` keyword.  
They can also be used as "Dynamic values" along side data from Data Scripts. In a component, Repeaters/Collections can use array/object std. props (discussed below), and in such cases all children of repeater/collection block receive default props named "item" in case of array or "key" and  "value in case of objects. These default props behave as any other prop and thus can also in turn be used as Dynamic value.
 

https://github.com/user-attachments/assets/abc54c12-afb6-477d-bb5f-82a83b2ed906

This addition brings some of the benefits of component-based frameworks into the builder.
By associating scripts and props directly with blocks (and therefore components), it becomes easier to create **reusable components** and **component libraries**.

#### Standard Props for Components:
Based on block props and block scripts, this PR also introduces "Standard Props for Components". Root block of each component can have special block props called "Standard Props". These props act as an interface between the component and the user. These props can be of various types (unlike normal block props which are always of type string) such as string, number, boolean, select, array and object. When a component with std. props is used in a page, the user gets a form where he/she can fill out the value for those std. props.
For eg:
- "Profile Card" component can have std. props like `Name`, `Description`, `Image URL`, etc. When this component is used in a page, the user simply fills those inputs and the Profile Card is ready.
Screenshot 2025-11-26 at 11 05 25 PM

https://github.com/user-attachments/assets/6cef3069-d62e-498f-9ee4-52dbe5b1590f

- "Navbar" component can have a std. prop named `Links` in which user puts a list of title and urls.

https://github.com/user-attachments/assets/7d3411c3-cc10-42fa-9f0a-cd3bd59f3513

- "Weather Card" can take just the location and show details automatically

https://github.com/user-attachments/assets/08664bbf-3627-4129-ae64-8531e2eca360

### DIFF
--- a/builder/builder/doctype/builder_page/builder_page.py
+++ b/builder/builder/doctype/builder_page/builder_page.py
@@ -332,6 +332,12 @@ class BuilderPage(WebsiteGenerator):
 			return True
 
 	def set_style_and_script(self, context):
+		builder_settings = frappe.get_cached_doc("Builder Settings", "Builder Settings")
+		if builder_settings.script:
+			context.setdefault("scripts", []).append(builder_settings.script_public_url)
+		if builder_settings.style:
+			context.setdefault("styles", []).append(builder_settings.style_public_url)
+
 		client_scripts = self.get("client_scripts") or []
 		for script in client_scripts:
 			script_doc = frappe.get_cached_doc("Builder Client Script", script.builder_script)
@@ -346,21 +352,15 @@ class BuilderPage(WebsiteGenerator):
 		if not context.get("_body_html"):
 			context._body_html = ""
 
-		if self.head_html:
-			context._head_html += self.head_html
-
-		if self.body_html:
-			context._body_html += self.body_html
-
-		builder_settings = frappe.get_cached_doc("Builder Settings", "Builder Settings")
-		if builder_settings.script:
-			context.setdefault("scripts", []).append(builder_settings.script_public_url)
-		if builder_settings.style:
-			context.setdefault("styles", []).append(builder_settings.style_public_url)
 		if builder_settings.head_html:
 			context._head_html += builder_settings.head_html
 		if builder_settings.body_html:
 			context._body_html += builder_settings.body_html
+
+		if self.head_html:
+			context._head_html += self.head_html
+		if self.body_html:
+			context._body_html += self.body_html
 
 		context["_head_html"] = render_template(context._head_html, context)
 		context["_body_html"] = render_template(context._body_html, context)
--- a/.github/workflows/on_release.yml
+++ b/.github/workflows/on_release.yml
@@ -15,9 +15,9 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
       - name: Setup Node.js
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
       - name: Setup dependencies
         run: |
           npm install @semantic-release/git @semantic-release/exec --no-save
--- a/.github/workflows/server-tests.yml
+++ b/.github/workflows/server-tests.yml
@@ -41,14 +41,14 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v6
         with:
-          python-version: '3.10'
+          python-version: '3.14'
 
       - name: Setup Node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 24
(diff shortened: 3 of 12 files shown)


stravo1:
Remaining features:

- [ ] Standard props in Component Root Block
- [ ] Reuse Block Scripts for Components instead of creating extra `<script>` tags
- [ ] Autocorrect in Block Script for "this" and "props"

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -488,10 +489,41 @@ def get_block_html(blocks):
...
+			props_obj_js = "{}" # to be passed as the prop obj in block script
+			props_with_successors = []
+   
+			if block.get("props"):
+       
+				# collect all key:value pairs
+				props_js = []
+				props_jinja = []
+    
+				for key, value in block.get("props", {}).items():
```
stravo1: use prop_name, prop_value for better readability

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -800,6 +869,20 @@ def resolve_path(path):
 	return original_resolve_path(path)
 
 
+def get_interpreted_prop_value(prop, data_key, map_of_inherited_props):
+	prop_value = prop["value"]
+	prop_type = prop["type"]
+
+	if prop_value == "" or prop_value is None:
+		return "undefined"
```
stravo1: empty string should not be interpreted as undefined

frontend/src/components/BlockProperties.vue
```diff
@@ -102,6 +120,22 @@ const getFilteredProperties = (section: PropertySection) => {
 	});
 };
 
+const getFilteredPropertiesDynamic = (section: PropertySectionDynamic) => {
```
stravo1: unify with existing "static" functions?

frontend/src/components/BlockPropertySections/StandardPropsInputSection.ts
```diff
@@ -0,0 +1,103 @@
...
+			getProps,
+			searchKeyWords: propKey,
+		});
+	}
+	console.log("standard props sections: ", sections);
+	return sections;
+};
+
+export default {
+	name: "Standard Props",
```
stravo1: Change section name

frontend/src/components/BuilderBlock.vue
```diff
@@ -116,19 +116,33 @@ const attributes = computed(() => {
...
 		props.block.dynamicValues
 			?.filter((dataKeyObj: BlockDataKey) => {
 				return dataKeyObj.type === "attribute";
 			})
 			?.forEach((dataKeyObj: BlockDataKey) => {
 				const property = dataKeyObj.property as string;
-				attribs[property] =
-					getDataForKey(props.data as Object, dataKeyObj.key as string) ?? attribs[property];
+				let value;
+				if (dataKeyObj.comesFrom === "dataScript") {
```
stravo1: same here, just like `frontend/src/components/BlockHTML.vue`, we need check `comesFrom == "prop"` first

frontend/src/components/PropsDependencyEditor.vue
```diff
@@ -0,0 +1,288 @@
...
+				:make-fixed="true"
+				fix-to=".props-popover-content"
+				class="w-full [&>.form-input]:border-none [&>.form-input]:hover:border-none [&>div>input]:text-xs"
+				v-bind="events"
+				ref="autoCompleteRef"
+				placeholder="Prop Name"
+				:modelValue="key"
+				:getOptions="getAutocompleteOptions"
+				@update:modelValue="
+					(option) => {
```
stravo1: can be extracted out to a handler func?

Issue #310, opened 2025-02-12
[FEATURE] Change ARIA Properties via the Builder Sidebar


harshtandiya:
[ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) is a standard for web accessibility.

I think a very valuable addition to builder would be to add an `Accessibility` section somewhere the right side control panel which would enable the developer to define aria labels for components with ease.

Right now, the only way to do it is via "HTML Attributes"

VaideshWaranR:
Hey, is someone already working on this issue or can I take it? @surajshetty3416

aerodeval:
Here is an example implementation  for the above.

![Image](https://github.com/user-attachments/assets/408c4a01-5ffa-45d0-999d-f63bd3e71819)

Please let me know if you have any feedback or suggestions for improvement.

surajshetty3416:
https://github.com/frappe/builder/pull/431

PR #431, opened 2025-11-15
feat: Add Accessibility section 

aerodeval:
Changes:

This PR introduces a new Accessibility section in the Builder sidebar that allows users to modify ARIA properties directly from the UI. 

Currently it supports updating ARIA-related attributes such as:

- Label Text - for adding custom aria-label.
- Role - dropdown with common options like button, navigation, banner, main, etc. to override semantic UI depending on the use case.
- Tab Index - to manage keyboard focus order (would be -1, 0 or 1).

This update will help users to configure ARIA attributes visually, making the Builder more inclusive and easier to use for teams focused on accessibility standards.

implements [feature 310](https://github.com/frappe/builder/issues/310)

frappe-aria

### DIFF
--- a/frontend/src/components/BlockProperties.vue
+++ b/frontend/src/components/BlockProperties.vue
@@ -32,6 +32,7 @@
 	</div>
 </template>
 <script setup lang="ts">
+import accessibilitySection from "@/components/BlockPropertySections/AccessibilitySection";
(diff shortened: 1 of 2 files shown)


surajshetty3416 (changes requested):
@aerodeval  can you move [Tag](https://github.com/frappe/builder/blob/develop/frontend/src/components/BlockPropertySections/OptionsSection.ts#L13) property inside the Accessibility section?

PR #446, opened 2025-12-17
fix(Code Editor): tab behaviour on multiline selection

ruchamahabal:
**Before**:

Selecting multiple lines and pressing tab deletes the block

https://github.com/user-attachments/assets/452997fa-f1e9-435b-a064-e9e383ad35a9

**After**:

Use codemirror's `indentMore`, `indentLess` commands to indent on tab, and deindent on shift + tab

https://github.com/user-attachments/assets/639814d1-ce25-4c3b-be6d-3d26de1b3235

### DIFF
--- a/frontend/src/utils/createCodeMirrorState.ts
+++ b/frontend/src/utils/createCodeMirrorState.ts
@@ -8,6 +8,8 @@ import {
 	defaultKeymap,
 	history,
 	historyKeymap,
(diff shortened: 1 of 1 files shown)


NagariaHussain:
Code Editor component should be moved to Frappe UI? 👀

surajshetty3416 (approved):

PR #447, opened 2025-12-18
fix: validate color input and prevent invalid hex values

### LINKED ISSUE #434: Color field accepts incorrect hex values
Image 

The first one should not be allowed.

Vibhuti410:
## Issue:
The color input field accepted invalid values like incorrect or incomplete hex values. On pressing Enter, the invalid values were accepted without any verification feedback.

## Solution:
Added validation for color input field.
Now it shows visual error if invalid values are entered.
Prevents appplying invalid hex values.

## After:
https://github.com/user-attachments/assets/18812778-67d1-4b12-8013-f08f8d939370

Note: This PR focuses on hex and functional color formats. Support for named CSS colors (e.g 'red' , 'black') can be added as separate follow up improvement.

Closes #434

### DIFF
--- a/frontend/src/components/Controls/ColorInput.vue
+++ b/frontend/src/components/Controls/ColorInput.vue
@@ -38,37 +38,7 @@
 											}
 										: undefined
 								"
-								@update:modelValue="
-									(val: string | null) => {
-										//auto strip extra hashes
-										if (typeof val == 'string') {
-											val = normalizeColorInput(val);
-										}
-
-										//last valid value is retained
-										const isInvalid = !!val && !isValidColorInput(val);
-										if (isInvalid) {
-											
-											const lastValidValue = props.modelValue;
-											// step 1: clear input
-											emit('update:modelValue', null);
-
-											// step 2: restore last valid value after render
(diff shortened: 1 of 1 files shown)


Vibhuti410:
@surajshetty3416 Thank you for your suggestion.

The current logic already keeps the last valid input and doesn't apply invalid changes. I agree the changes you suggested would result in a more intuitive UX. I will look into it and try to make the required changes.

Vibhuti410:
@surajshetty3416 As per your suggestion I have updated the implementation to auto-strip the extra hashes if the rest of the input is valid. It also avoids applying invalid values along with visually highlighting the incorrect input with red border and retains the last valid input color.

# Demo:

https://github.com/user-attachments/assets/8b75a396-b1c4-4e1e-9770-43666b563ace

Vibhuti410:
@surajshetty3416 

https://github.com/user-attachments/assets/79eb7d71-17ec-413b-8737-0f4b5699f587

Removed the red outline. When an invalid input is entered, it is not applied and the last valid color is retained. The auto-strip extra hashes logic remains unchanged.

PR #459, opened 2026-01-06
feat: add block data and client scripts and block props

stravo1:
This PR introduces **block props** and **block client scripts (JS)** and **block data scripts (Py)**.

Each block now has three new properties: **Block Props** and **Block Client Scripts** and **Block Data Scripts**.

* **Props:**
  Props work similarly to those found in component-based JavaScript frameworks such as Vue or React. They serve as an interface between the “outside world” and the block scripts (both client and data scripts).

  There are two types of props:

  * **static** - Constant literal values.
  * **dynamic** - Derived from Data Scripts (both Page Data and Block Data).
  
  Props are **not passed down to child blocks by default**. However, users can explicitly choose which props should be propagated. Any prop that is marked to be passed down becomes accessible to **all descendant blocks** of the block that declares it. These props can also be used as **dynamic values** wherever dynamic values are supported, such as in Text blocks.
  
  **Standard Block Props**
        
    These are special type of block props which can be used in Components. Std. Block props can be **declared on the root block of a component.** They can be of various types such as string, number, boolean, arrays (`string[]`) and objects (`Map<string, string>`), unlike non-std. props which are always of type string. They differ from from non-std. props in two primary ways:

     - When a component is used in a Page, the **std. props appear as user inputs on the Right Panel** (provided the root block of the component is selected). Different types of std. props will show up as different inputs. This way different instances of a component can take different inputs and produce different results. Std. Props **can also have default values** which are used if the user does not provide any input to the component using std. prop inputs. Non-std. props are like variables which are declared and defined at the same time, while std. props are like parameters of a function (here the function is a component) which can take different values but can also fallback on default values if not provided.
     - Std. props are passed down by default.
     
    **Std. props of type array and object can also be used as keys for Repeaters/Collections** to loop over their values. In such cases, for std. prop of type object, all blocks - which are repeated - are assigned two _default props_ named `key` and `value` which hold the value of the key-value pairs of the object being looped over. Same goes for std. prop of type array, and in such case there is only one _default prop_ named `item`.

Adding block props:

Screenshot 2026-01-06 at 11 01 11 AM

Adding std. props to root of component:

Screenshot 2026-01-06 at 10 41 29 AM

Different types of std. props:

Screenshot 2026-01-06 at 10 41 40 AM

Some std. props with default values (here Name and Type are same, that is for demonstration purpose) :

Screenshot 2026-01-06 at 10 39 53 AM

How they appear when used in a page:

Screenshot 2026-01-06 at 10 54 41 AM

* **Block Data Script:**
  These are Python code snippets associated with an individual block, unlike Data Script which applies to the entire page. Block data scripts can be used like Page Data scripts to get dynamic values during SSR.

  Within Block Data Script, two special keywords are available:

  * `block`: Similar to `data` in page data scripts, we have `block` in block data scripts. It can be used to assign dynamic values which can be used in the block. Also **all block data are passed down**, so once a block assigns a dynamic value, it is accessible to all its successors. A successor block can re-assign a dynamic value in it's own block data script in which case the nearest assignment of the value is used, similar to how block scoping/variable shadowing works.

    ```py
    block.sample_dynamic_value = "Hello world!"
    ```

  * `props`: Provides access to the properties passed to the block.

Assigning block data script:

Screenshot 2026-01-06 at 10 58 31 AM

Screenshot 2026-01-06 at 10 58 37 AM

Screenshot 2026-01-06 at 10 58 49 AM

Using the cumulative toggle to view all the block data being passed to the current block.

Screenshot 2026-01-06 at 10 59 01 AM

 
* **Block Client Script:**
  These are JavaScript code snippets associated with an individual block, unlike **Client Scripts** which apply to an entire page.
  Block Scripts make it easier to manage reactivity by eliminating the need for multiple Client Scripts and by localizing logic to the block itself.

  Within a Block Script, two special keywords are available:

  * `this`: Refers directly to the DOM node corresponding to the block.
    Example:

    ```js
    this.addEventListener('click', ...)
    ```

    attaches an event listener to the block itself.
  * `props`: Provides access to the properties passed to the block.
  

Some examples:
- "Item Card" takes only the item name and prepares the entire item card (SSR):

https://github.com/user-attachments/assets/cb97b93c-e4c5-480a-8be3-90d56e562a93

- "Profile Card" component can have std. props like `Name`, `Description`, `Image URL`, etc. When this component is used in a page, the user simply fills those inputs and the Profile Card is ready.

https://github.com/user-attachments/assets/6cef3069-d62e-498f-9ee4-52dbe5b1590f

- "Navbar" component can have a std. prop named `Links` in which user puts a list of title and urls.

https://github.com/user-attachments/assets/7d3411c3-cc10-42fa-9f0a-cd3bd59f3513

- "Weather Card" can take just the location and show details automatically

https://github.com/user-attachments/assets/08664bbf-3627-4129-ae64-8531e2eca360

### DIFF
--- a/frontend/src/components/BlockProperties.vue
+++ b/frontend/src/components/BlockProperties.vue
@@ -14,7 +14,7 @@
 				</template>
 			</CollapsibleSection>
 		</div>
-		<div class="sticky top-0 mt-[-16px] flex w-full bg-surface-white py-3">
+		<div class="sticky top-[1px] z-[1] mt-[-16px] flex w-full bg-surface-white py-3">
 			<BuilderInput
 				ref="searchInput"
 				type="text"
--- a/frontend/src/components/BuilderLeftPanel.vue
+++ b/frontend/src/components/BuilderLeftPanel.vue
@@ -3,6 +3,7 @@
 		<PanelResizer
 			:dimension="builderStore.builderLayout.leftPanelWidth"
 			side="right"
+			:minDimension="200"
 			:maxDimension="500"
 			@resize="(width) => (builderStore.builderLayout.leftPanelWidth = width)" />
 		<div
--- a/frontend/src/components/DashboardSidebar.vue
+++ b/frontend/src/components/DashboardSidebar.vue
@@ -116,7 +116,7 @@
 					),
 				}"
 				@click="setFolderActive(project.folder_name)">
-				<span class="flex flex-1 gap-2">
+				<span class="flex flex-1 gap-2 overflow-hidden">
 					<FolderIcon class="size-4"></FolderIcon>
 					<EditableSpan
 						v-model="project.folder_name"
--- a/frontend/src/components/EditableSpan.vue
+++ b/frontend/src/components/EditableSpan.vue
@@ -1,6 +1,7 @@
 <template>
 	<div
 		ref="editableRef"
+		class="overflow-hidden truncate rounded-sm px-1 focus-within:border focus-within:border-outline-gray-2 focus-visible:outline-none"
 		:contenteditable="editMode"
 		@dblclick="handleDoubleClick"
 		@blur="handleBlur"
--- a/frontend/src/components/PageOptions.vue
+++ b/frontend/src/components/PageOptions.vue
@@ -7,17 +7,18 @@
 				class="w-full text-sm [&>label]:w-[60%] [&>label]:min-w-[180px]"
 				:modelValue="page.page_title"
 				:disabled="builderStore.readOnlyMode"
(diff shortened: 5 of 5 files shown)


builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -850,5 +1046,85 @@ def reset_block(block):
...
+				return None
+		case "object":
+			try:
+				return frappe.parse_json(value)
+			except Exception:
+				return None
+		case "select":
+			return f"{value}"
+		case _:
+			return f"{value}"
```
surajshetty3416: could be simplified? also moved to utils?

NagariaHussain:
Awesome 🙌

Issue #460, opened 2026-01-10
Absence of dynamic values in component used in a page results in unparsed Jinja string


stravo1:
#### Steps to reproduce:
1. Create a text block in a container.
2. Save a data script as in:
```py
data.details  = {
  "title": "hello world!"
}
```
3. Assign the text block the dynamic value for the text block as details.title
4. Save the container as Component
5. Change the data script to alter the key, so maybe change “title” to “titlee” to replicate absence of the dynamic value.
6. Although in Canvas you can see the correct fallback text, when you publish you see an unparsed Jinja string.

https://github.com/user-attachments/assets/c82ae265-39a3-4b2d-a1c9-36efff779003

surajshetty3416:
https://github.com/frappe/builder/pull/461

PR #461, opened 2026-01-10
fix(dynamicValues): unparsed Jinja string in absence of dynamic values in components

stravo1:
solves #460 

the issue occurred because resetting blocks does not remove dynamic values, so when a component is used in a page, all it's dynamic values are replicated into the block from the component. now while rendering the page, a block's dynamic values are calculated by combining the dynamic values of both the block and the reference block from the component from which it is derived and thus it results in replicated dynamic values. and while resolving the dynamic values the fallback value is recalculated twice and the last time it uses the result which was calculated previously.

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -42,9 +42,10 @@ jobs:
         with:
           python-version: '3.14'
 
-      - uses: actions/setup-node@v3
+      - name: Setup Node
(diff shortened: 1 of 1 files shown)


surajshetty3416:
@stravo1 Let's add a couple of backend tests for this?

stravo1:
> @stravo1 Let's add a couple of backend tests for this?

sure 👍🏼

Issue #465, opened 2026-01-24
Save Component dialog does not respond to enter Key


Hemant-dbit:
When saving a component, pressing Enter in the component name field doesn't save - users must click the Save button with mouse.

PR #466, opened 2026-01-24
fix: Allow Enter key to save Component

Hemant-dbit:
**Fixes: #465**

**Issue:**
Saving a component required clicking the Save button with a mouse. Pressing Enter in the component name field had no effect

**Solution:**
- Added @keydown.enter.prevent event handler to the component name input field
- Created handleEnterKey() function to validate input and trigger save action
- Added [v-model="showDialog"] to properly control dialog visibility

https://github.com/user-attachments/assets/c97b1c33-7598-47e0-b3f9-ca19d68b7dad

@surajshetty3416

### DIFF
--- a/frontend/src/components/Modals/NewComponent.vue
+++ b/frontend/src/components/Modals/NewComponent.vue
@@ -17,7 +17,7 @@
 				type="text"
 				:modelValue="componentName"
 				@input="(value: string) => (componentName = value)"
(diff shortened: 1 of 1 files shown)


Hemant-dbit:
Thanks for the feedback! 
I agree that global Enter-to-save could be unexpected.

PR #507, opened 2026-03-04
feat: Sync `builder_files` with page/script delete and rename operations

### LINKED ISSUE #496: bug: Sync export to app
builder 1.22.4 master (5377f0a)

The "Make this page a standard page" crea the file on source code of a custom app, but if you delete this page, delete client script js or CSS it only remove it from db not from source code, so looking yo the source code of the custom app you will see a lot of delete residual file of deleted page/scripts

Steps

1. Create a new page and make it standard page of you custom app
2. Clone it
3. Delete the cloned page
4. Go to apps/custom_app/custom_app/builder_files/pages, you will see 2 folders but on the frappe builder UI you only have 1 page

This should be like frappe Doctype, if you delete it on the UI it also delete it from the code.

asaura08:
### Problem

When a page is marked as a standard page, Builder writes its source files to
`{app}/builder_files/pages/{page_name}/`. However, when that page (or any of its
client scripts) was deleted or renamed through the Builder UI or Desk, the
corresponding files on disk were **not** removed or updated. This left orphaned
directories in the app's source code that had no corresponding record in the
database — inconsistent with how Frappe handles standard DocTypes in developer
mode.

fix #496 

### Solution

Mirrors the behaviour of Frappe's own DocType sync in developer mode: any
filesystem artifact created when a record is exported must be cleaned up when
that record is deleted or renamed.

now
image
image

#### Changes

**`builder/export_import_standard_page.py`**
- `delete_standard_page_files(page_name, app_name)` — removes
  `builder_files/pages/{export_name}/`.
- `delete_standard_client_script_files(script_name, app_name)` — removes
  `builder_files/client_scripts/{export_name}/`.
- `rename_standard_client_script_files(old, new, app_name)` — renames the
  client-script directory and its inner JSON file.

**`builder/doctype/builder_page/builder_page.py`**
- `on_trash` — when `is_standard`, `app` and `developer_mode` are set, deletes
  the page directory and any client-script directories that are no longer
  referenced by another standard page in the same app.
- `after_rename` — deletes the old page directory; the subsequent `on_update`
  call re-exports the page under the new name automatically.

**`builder/doctype/builder_client_script/builder_client_script.py`**
- `on_trash` — removes the script's `builder_files` directory for every app
  whose standard pages reference it.
- `after_rename` — renames the script directory in every referencing app.
- `_get_referencing_apps()` — helper that returns the unique set of apps whose
  standard pages link to this script.

**`frontend/src/components/Settings/PageGeneral.vue`**
- Added a **Page Name** field (visible in developer mode when the page is
  standard) that allows renaming the exported file directly from the Builder UI,
  without having to go to Desk.
- The rename is committed on `Enter` or when the field loses focus (`focusout`).

**`builder/doctype/builder_page/builder_page.js`**
- Fixed a bug where the installed-apps list was being assigned to the wrong
  field (`module` instead of `app`), which caused the **App** select in the
  Desk form to appear empty.

**`frontend/src/components/PageActionsDropdown.vue`**
- The **Delete** action is now shown for standard pages when `developer_mode` is
  active, consistent with the rest of the developer-mode affordances.

#### Tests

`TestStandardPageSync` (6 cases) covers:
1. `on_trash` removes the page directory.
2. `on_trash` removes an orphaned client-script directory.
3. `on_trash` does **not** remove a client-script directory shared by another
   standard page.
4. `after_rename` deletes the old directory and re-exports under the new name.
5. Client-script `on_trash` removes the script's directory.
6. Client-script `after_rename` renames the directory and inner JSON file.


stravo1:
Hi, thanks for the PR!

I noticed a couple of things that could be improved. It works correctly when directly deleting a standard page (more on this below), but if I first uncheck “Standard Page,” the files remain. Then, when deleting that now non-standard page, the files are still not removed, since `on_trash` checks the `is_standard` property, which is now false.

In the PR, deleting a standard page is allowed in Developer Mode. However, I feel it might be better to keep the “Delete” action in `PageActionsDropdown.vue` hidden for standard pages, even in Developer Mode.

Perhaps a cleaner approach would be to delete the page files when the “Standard Page” toggle is unchecked. This would also remove the need to enable the “Delete” action in Developer Mode.

@asaura08 @surajshetty3416

surajshetty3416:
@stravo1 Can you complete the review and merge?

@asaura08 you might have to resolve the conflicts.

surajshetty3416:
> @stravo1 Can you complete the review and merge?

@stravo1 any update?

builder/builder/doctype/builder_client_script/builder_client_script.py
```diff
@@ -39,6 +39,39 @@ def on_update(self):
...
+
+	def _get_referencing_apps(self) -> list[str]:
+		"""Return the distinct app names of standard pages that use this script."""
+		refs = frappe.get_all(
+			"Builder Page Client Script",
+			filters={"builder_script": self.name},
+			fields=["parent"],
+			ignore_permissions=True,
+		)
+		apps: list[str] = []
```
stravo1: Can we not use a single set only (and not the array)?

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -196,11 +196,44 @@ def on_update(self):
...
 
+		if self.has_value_changed("is_standard") and not self.is_standard:
+			self._delete_standard_page_files_if_needed()
+
 	def clear_route_cache(self):
 		get_web_pages_with_dynamic_routes.clear_cache()
 		find_page_with_path.clear_cache()
 		clear_cache(self.route)
 
+	def _delete_standard_page_files_if_needed(self) -> None:
```
stravo1: rename to `_cleanup_standard_page_exports` ?

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -196,11 +196,44 @@ def on_update(self):
...
+
+		from builder.export_import_standard_page import (
+			delete_standard_client_script_files,
+			delete_standard_page_files,
+		)
+
+		delete_standard_page_files(self.page_name, app)
+
+		client_scripts = doc_before.client_scripts if doc_before else self.client_scripts
+		for row in client_scripts:
```
stravo1: something like
```python
        for row in client_scripts:
			referencing_apps = (
				frappe.get_doc("Builder Client Script", row.builder_script)
			)._get_referencing_apps()

			if app not in referencing_apps:
                ....
```
so that we re-use code which is already written and also reduce LOC here

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -212,6 +245,20 @@ def on_trash(self):
...
+
+	def after_rename(self, old: str, new: str, merge: bool = False) -> None:
+		if not (self.is_standard and self.app and frappe.conf.developer_mode):
+			return
+		from builder.export_import_standard_page import (
+			delete_standard_page_files,
+			export_page_as_standard,
+		)
+
+		delete_standard_page_files(old, self.app)
```
stravo1, line 318: why not rename the files like we are doing for client scripts?

builder/export_import_standard_page.py
```diff
@@ -315,6 +315,65 @@ def export_variables(variables, builder_files_path):
...
+	app_path = frappe.get_app_path(app_name)
+	if not app_path:
+		return
+	export_name = frappe.scrub(script_name)
+	script_path = os.path.join(app_path, "builder_files", "client_scripts", export_name)
+	if os.path.isdir(script_path):
+		shutil.rmtree(script_path, ignore_errors=True)
+
+
+def rename_standard_page_files(old_page_name: str, app_name: str) -> None:
```
stravo1: This is unused, maybe we can re-write this to handle the renaming of the std. page files?

Am I missing something? Is deleting and re-exporting better in some manner?

frontend/src/components/Settings/PageGeneral.vue
```diff
@@ -152,6 +152,26 @@
 									@update:modelValue="handleAppChange"></BuilderInput>
 							</div>
 						</div>
+						<template v-if="pageStore.activePage?.is_standard && isDeveloperMode">
```
stravo1: @surajshetty3416 is this needed? maybe handy to not have to visit Desk to rename pages

stravo1:
Also if I am not wrong, components are not exported automatically when saved. To make sure that they are updated you need to edit a std. page which uses that component. We can solve this the way we are tackling client scripts?

@surajshetty3416 @asaura08

builder/builder/doctype/builder_page/builder_page.py
stravo1: Although this increases one DB visit.

@surajshetty3416 opinions?

asaura08: Good call! I've refactored the loop to reuse `_get_referencing_apps()` as suggested.

One thing to note: since we're calling this during a delete or uncheck of the current page, `_get_referencing_apps()` could still return the current page's app if it hasn't been removed from the DB yet — which would incorrectly prevent the script files from being deleted. To handle this, I've added an optional `exclude_page` parameter to `_get_referencing_apps()` so we can exclude `self.name` from the lookup:

```python
for row in client_scripts:
    referencing_apps = frappe.get_doc(
        "Builder Client Script", row.builder_script
    )._get_referencing_apps(exclude_page=self.name)

    if app not in referencing_apps:
        delete_standard_client_script_files(row.builder_script, app)
```

This does add one extra DB query per script (a `get_doc` on top of the existing `get_all`), but it avoids duplicating the logic. Happy to keep the inline approach if you'd prefer to avoid the extra query.

builder/builder/doctype/builder_page/builder_page.py
asaura08: Good point! I've added a `rename_standard_page_files(old, new, app_name)` function to `export_import_standard_page.py`, mirroring the existing `rename_standard_client_script_files`. It renames the page directory and its inner JSON file in one step.

`after_rename` now simply calls it:
```python
def after_rename(self, old: str, new: str, merge: bool = False) -> None:
    if not (self.is_standard and self.app and frappe.conf.developer_mode):
        return
    from builder.export_import_standard_page import rename_standard_page_files
    rename_standard_page_files(old, new, self.app)
```

I've also updated the corresponding test to mock `rename_standard_page_files` instead of the old `delete + export` pair.

stravo1:
> components are not exported automatically when saved

fixed

frontend/src/components/Settings/PageGeneral.vue
stravo1: removed

stravo1:
Of the +900 LOC, ~600 LOC is in a test file.

stravo1 (approved):

PR #512, opened 2026-03-05
test: Init playwright test setup

surajshetty3416:
- Replace cypress test setup with playwright for e2e tests


surajshetty3416:
Will open when am working on this actively

PR #515, opened 2026-03-12
feat: add keyboard shortcuts overlay

yohaann196:
Adds a keyboard shortcuts reference overlay that can be toggled by pressing '?' in the editor and it displays all existing shortcuts grouped by category (like tools, canvas, editing, panels) and it closes with Esc or also thru clicking the backdrop. No existing behaviour is changed


surajshetty3416:
hey @yohaann196, nice one!

But I have been wanting to implement this in more generic and scalable way, so raised this: https://github.com/frappe/builder/pull/516

With this we just have to mention group and description where you setup the shortcut, `useShortcut` composable will handle the rest... you don't have to maintain a separate shortcut map to show it in a modal.

closing this for now

PR #519, opened 2026-03-12
feat: Style presets

### LINKED ISSUE #106: feat: style book
It would be really useful if we could define global styling for basic HTML elements globally (and/or based on an additional "Builder Theme" abstraction layer).

I suggest that this include a pre-set collection of blocks, including, but not limited to:

- [ ] Headings (h1-h7)
- [ ] Paragraph
- [ ] Link
- [ ] Lists (ol & ul)
- [ ] Quote
- [ ] Table

Some of these needs can be covered by global components and/or custom stylesheets, but it's a bit tedious to create global components for every one of these, and style sheets don't currently load on the backend (and aren't visual).

Here's the WordPress Style Book from the Site Editor, for inspiration:
image

Vibhuti410:
Closes #106

## Changes
- Added `Builder Style Preset` DocType to manage presets from the desk with `style_name`, `style_map` (JSON) and `sort_order` fields
- Added a Style select field in the Typography panel to apply the presets on text blocks
- Fields that are controlled by the preset shows the value as a placeholder (inherited)
- Also can manually override a field and when we clear that, the fields automatically takes back the preset value

## New Files Added
- `stylePreset.ts` — added this to fetch all style presets from the DocType
- `presetUtils.ts` — added this to get the active preset's style map to avoid repetition and can be just imported wherever useful

## Changes In Existing Files
- `TypographySection.ts` — added the Style preset dropdown here
- `StylePropertyControl.vue` — extended this to handle preset inheritence for all style fields
- `blockController.ts` — added `getPresetStyle` and `setPresetStyle` methods here
- `block.ts` — added `presetStyle` property to keep track of which preset is applied on a block

## Demo:
https://github.com/user-attachments/assets/515b810f-6edc-4835-93f5-664193333ba8


frontend/src/components/BlockPropertySections/TypographySection.ts
```diff
@@ -34,6 +43,54 @@ const typographySectionProperties = [
...
+					if (!val) {
+						blockController.setPresetStyle("");
+						Object.values(styleKeyMap).forEach((cssProperty) => {
+            				blockController.setStyle(cssProperty as styleProperty, null);
+          	        });
+        			return;
+					}
+					Object.values(styleKeyMap).forEach((cssProperty) => {
+					blockController.setStyle(cssProperty as styleProperty, null);
+					});
```
surajshetty3416: Handle this style mapping inside BuilderBlock.vue where we compute style map

surajshetty3416 (changes requested):

frontend/src/components/BlockPropertySections/TypographySection.ts
```diff
@@ -81,7 +109,6 @@ const typographySectionProperties = [
 				actionButton: {
 					component: FontUploader,
 				},
-				getModelValue: () => blockController.getFontFamily(),
```
surajshetty3416: why 🤔

surajshetty3416 (changes requested):

surajshetty3416 (changes requested):
@Vibhuti410 you need to ship the standard style presets json files as well so that users get predefined styles on migrate.

check this 
https://github.com/frappe/builder/pull/379/changes/af56679bef541258cc61a49dac5867bc39a36b1e

and this for reference
https://github.com/frappe/builder/pull/379/changes/d844000609b7766cd122ff84ba365eb767b36c37

surajshetty3416:
@Vibhuti410 Please reopen once this is ready and all the conflicts are resolved.

PR #540, opened 2026-03-31
fix: replaced `prev_blocks` with `block` for inherited block data

stravo1:
now all blocks can access all prev blocks' data using the `block` keyword itself

ref: https://github.com/frappe/builder/pull/533

new simplified process:
Screenshot 2026-03-31 at 5 49 49 PM


surajshetty3416 (approved):

surajshetty3416:
@mergifyio backport master

PR #541, opened 2026-03-31
fix: select dropdown rendering as text input in overflow

### LINKED ISSUE #501: Overflow x is a free text field instead of dropdown
Image

Vibhuti410:
Declared types and options as props so it reaches the input component and renders the dropdwon correctly.

Before:

https://github.com/user-attachments/assets/4cfd902f-57b4-4e83-b974-ba7eaa5af00d

After:

https://github.com/user-attachments/assets/f94fa339-df2f-4a7e-bdca-55ff3bdd88da

Fixes: #501


surajshetty3416 (approved):

surajshetty3416:
@Mergify update

PR #547, opened 2026-04-12
feat: Increment/Decremet arrow keys

### LINKED ISSUE #546: Increment/Decrement arrow keys
Add increment/decrement arrows on hover for numeric input.

Vibhuti410:
Added arrows keys to increment/decrement the numeric input on hover.

Demo:

https://github.com/user-attachments/assets/fc1c51f6-3812-4483-a491-3d48a2cf8b9c

Closes #546


surajshetty3416 (commented):
@Vibhuti410 

- [ ] The gap between arrows should be less so that it has some space around the edges (we already discussed this)
- [ ] For few inputs, arrow keys does not make sense or the step should be different... example, in font weight it can be incremented/decremented by 100 instead of 1... it should also have min and max constraints. Please go through all such inputs
	Screenshot 2026-04-13 at 1 07 57 PM

surajshetty3416 (changes requested):
@Vibhuti410 

- [x] The gap between arrows should be less so that it has some space around the edges (we already discussed this)
- [x] For few inputs, arrow keys does not make sense or the step should be different... example, in font weight it can be incremented/decremented by 100 instead of 1... it should also have min and max constraints. Please go through all such inputs
	Screenshot 2026-04-13 at 1 07 57 PM
- [ ] it should also auto detect the step of increment/decrement base on cursor position. 
		Screenshot 2026-04-13 at 1 13 57 PM
		In this case, it should increment decimal value and not the whole number

frontend/src/utils/useNumberInput.ts
```diff
@@ -0,0 +1,37 @@
+import { computed } from "vue";
```
surajshetty3416, line 1: [we use tab for indentation](https://github.com/frappe/builder/blob/develop/frontend/.prettierrc#L2)... you have used spaces. Please setup pre-commit and auto-formatter (in your editor) to avoid this

surajshetty3416 (changes requested):

surajshetty3416 (approved):
Resolved

surajshetty3416 (dismissed):
@Vibhuti410 
- Opacity is still that disc's opacity is still 0.7, we discussed it to make it 0.5 (i think we can go down to 0.3)
- Also, the "click and hold" functionality to continuously change the value is yet implemented

PR #561, opened 2026-04-24
fix(UX): Misc

### LINKED ISSUE #558: [UX] Unable to use arrow keys to change values + cursor disappears
https://github.com/user-attachments/assets/904ef08c-2b1d-493b-b7b7-7c92a65e3412

New update to numeric fields does not allow using up down arrow keys to change values. The manual buttons are too small to practically use for power users. 

Cursor also changes to a touch selector while clicking the buttons to adjust. This shouldn't happen as it is almost invisible to see it. Cursor should change to a pointer instead.

### LINKED ISSUE #559: Line height adjuster should increment in integers instead of decimals for %
https://github.com/user-attachments/assets/374f0f37-a764-407c-a3c2-b7b511857f04

If line height is given in percentages, it should not increment or decrement in decimals. Only if non-percentage values are used, then decimals make sense.

Vibhuti410:
Changed opacity of circular cursor for better visibility on hover of arrow keys 

Fixed arrow key up and down functionality using keyboard which broke because of https://github.com/frappe/builder/commit/b595337c37c89072a67b0af4ed6040913160b529

Closes #558 
Closes #559

### DIFF
--- a/frontend/src/components/BlockPropertySections/TypographySection.ts
+++ b/frontend/src/components/BlockPropertySections/TypographySection.ts
@@ -123,7 +123,7 @@ const typographySectionProperties = [
 			return {
 				label: "Height",
 				propertyKey: "lineHeight",
-				step: 0.1,
+				enableSlider: true,
 			};
 		},
 		searchKeyWords: "Font, Height, LineHeight, Line Height",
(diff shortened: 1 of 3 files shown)


surajshetty3416:
@Mergify update

surajshetty3416 (approved):

PR #563, opened 2026-04-24
fix: InlineInput padding

### LINKED ISSUE #562: UI: InlineInput has extra padding on right
All InlineInputs now have "suffix" for NumberArrows which adds padding to the right. However, NumberArrows appear only on hover for number inputs.

Image

Input fields which are short like here have extra unnecessary padding which hides information without showing the NumberArrows.

Image
Image

### Steps to reproduce the behavior:
1. Select a block
2. Open the Position section
3. Set to "Free"

stravo1:
InlineInput now renders "suffix" conditionally to avoid adding extra padding to the right where not needed.
NumberArrows are always visible if it is a number field to avoid inconsistency.

fixes: #562

### DIFF
--- a/frontend/src/components/Controls/Input.vue
+++ b/frontend/src/components/Controls/Input.vue
@@ -22,7 +22,7 @@
 			<template #prefix v-if="$slots.prefix">
 				<slot name="prefix" />
 			</template>
-			<template #suffix>
+			<template v-if="hasSuffix" #suffix>
 				<NumberArrows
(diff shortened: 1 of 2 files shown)


stravo1:
NumberArrows being visible all the time is not the best UI, but maybe this:
Screenshot From 2026-04-24 21-51-13
looks better than this:
Screenshot From 2026-04-24 21-33-30

PR #567, opened 2026-04-28
fix: added conditonal padding in InlineInput

Vibhuti410:
Before:
image

After:
image

Close: #563


surajshetty3416:
resolved with https://github.com/frappe/builder/pull/575

PR #576, opened 2026-05-05
fix: spacing placeholder shows 'unset' correctly

Hemant-dbit:
**Issue**
Padding and margin values show 0 or blank instead of unset for new components.

**Fix**
Updated spacing fallback logic so padding and margin match the other property controls.

Fixes #535

**Before**

https://github.com/user-attachments/assets/d7d1cf4d-b304-4b02-8628-b9c6658ca954

**After**

https://github.com/user-attachments/assets/2b1aaa08-1ac3-466d-9c34-cb9f968d7748

cc @surajshetty3416


Hemant-dbit:
Added visuals.
@surajshetty3416

surajshetty3416 (approved):

PR #590, opened 2026-05-10
fix(Block Props): better default value parsing in getPropValue

stravo1:
For standard props, the prop value needs to be parsed (for eg: "true" -> true for Boolean), till now default values were not parsed — only values which were set manually were being parsed. This PR fixes that.


surajshetty3416:
@stravo1 description?

stravo1:
> @stravo1 description?

added

surajshetty3416 (approved):

PR #592, opened 2026-05-10
feat(DynamicValueHandler): truncate middle of key instead of the end

stravo1:
In deeply nested dynamic values, truncating end of the word gives very little context (if not none). Dynamic Value keys can be truncated at the middle to preserve context (similar to how file managers truncate filenames/paths), we also show full key on hover.

Before:
Screenshot 2026-05-10 at 11 57 05 PM

After:
Screenshot 2026-05-10 at 11 59 36 PM

https://github.com/user-attachments/assets/b024932e-b195-48d4-9c4d-1ccf66fd4e9d


surajshetty3416 (changes requested):
@stravo1  `MiddleTruncate` instead (better defines functionality)? 

And support 'as' prop? 

`<MiddleTruncate as='code' text='Looooooooooong-text-of-an-image.jpg'>`

surajshetty3416 (approved):

PR #596, opened 2026-05-13
feat: set component as global by right clicking the component

pratikb64:
- Add "Set as gobal component" option in context menu, to convert local component into global
- Toggle code view mode between "block" and "page" automatically based on whether a block is selected or not.


surajshetty3416:
@pratikb64 this should not be the part of block context menu (also rarely used).

Maybe we could have different context menu for components that will available in components list.

Closing this for now

PR #598, opened 2026-05-14
fix(Routing): Prefer last publish time for resolving routes with conflicts

stravo1:
If more than one page has the same route asscoaited with them, the current implementation always favors the last "created" page, so even if another page (having the same route but created before) is edited and published that page is not used for rendering. This PR introduces and uses last published time for resolving the route.

Before:

https://github.com/user-attachments/assets/c8544680-a7f8-45c5-b9f7-9610e0f1ea71

After:

https://github.com/user-attachments/assets/08c15768-ff95-44aa-9506-857548b56a8d

### DIFF
--- a/builder/builder/doctype/builder_page/builder_page.json
+++ b/builder/builder/doctype/builder_page/builder_page.json
@@ -15,7 +15,7 @@
   "page_name",
   "route",
   "dynamic_route",
-  "last_published_time",
+  "published_at",
   "column_break_ymjg",
   "is_standard",
   "is_template",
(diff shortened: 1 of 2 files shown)


stravo1:
Better maybe if we keep track of last published time?

surajshetty3416:
> Better maybe if we keep track of last published time?

Yep, better to do this

stravo1:
> Yep, better to do this

Done ✅ 

@surajshetty3416

builder/builder/doctype/builder_page/builder_page.json
```diff
@@ -242,11 +244,16 @@
    "label": "App",
    "mandatory_depends_on": "is_standard",
    "no_copy": 1
+  },
+  {
+   "fieldname": "last_published_time",
+   "fieldtype": "Datetime",
+   "label": "Last Published Time"
```
surajshetty3416:
```suggestion
   "label": "Published At"
```

surajshetty3416 (changes requested):

surajshetty3416 (approved):

PR #603, opened 2026-05-19
fix: Traceback now provides block_id which caused the error

stravo1:
It was very difficult to track which block caused an error. We now pass `block_id` to the filename parameter of `execute_script` function to help traceback the block which is the source of the error.

Screenshot 2026-05-19 at 8 40 21 AM

Screenshot 2026-05-19 at 8 39 54 AM


surajshetty3416 (changes requested):
@stravo1 Add a simple test case for this?

surajshetty3416 (approved):

PR #619, opened 2026-06-02
feat: redirect from data script

pratikb64:
### What
Adds the ability to redirect a page visitor to any URL from within a page **Data Script**.

    # example
    redirect("/login")          # 302 by default
    redirect("/login", 301)   # optional custom status code

### Why
Data scripts already run server-side on every render and can branch on the
session/user, but there was no way to send the visitor elsewhere. This closes that gap with a one-line helper.

### How
- Injects a `redirect(location, http_status_code=302)` helper into the page
  data script namespace. When called during an actual page render it sets
  `frappe.local.flags.redirect_location` and raises `frappe.Redirect`, which the
  Frappe website renderer turns into a real HTTP redirect — entirely
  server-side

### Security
- `get_page_data` is split into a whitelisted wrapper (always
  `for_render=False`) and an internal `_get_page_data(for_render=True)` used
  only by the render path — so a client cannot trigger the redirect via the API.
- Pages with a `page_data_script` are now marked `no_cache`. A data script can
  branch on the user/session, so without this a render for one user could be
  cached and served to another, bypassing a conditional redirect (auth-bypass).
- Redirect targets come from author-written server-side Python (same trust
  level as any data script); external URLs are intentionally allowed.

### Tests
- `test_redirect_from_page_data_script`: asserts the rendered response is a
  302 to the target URL, and that the editor's `get_page_data()` call is a
  no-op returning empty data.

### DIFF
--- a/builder/builder/doctype/builder_page/builder_page.py
+++ b/builder/builder/doctype/builder_page/builder_page.py
@@ -262,7 +262,7 @@ class BuilderPage(WebsiteGenerator):
 				"Builder Settings", "Builder Settings", "disable_auto_dark_mode"
 			)
 
-		page_data = self.get_page_data()
+		page_data = self._get_page_data(for_render=True)
 		if page_data.get("title"):
 			context.title = page_data.get("page_title")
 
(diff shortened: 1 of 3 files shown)


surajshetty3416 (approved):

surajshetty3416:
@pratikb64 Update docs as well?

https://docs.frappe.io/builder/data-script

PR #627, opened 2026-06-10
feat: Component scripts

stravo1:
Block Data Scripts [have been removed](https://github.com/frappe/builder/pull/623) for the following reasons:
- It was hard to discover.
- Code became too fragmented very quickly. In many cases, it became harder to manage than Page Scripts — the very problem it was trying to solve.
- Complicated entire codebase, hard to debug

> ⚠️ For similar reasons, this PR removes Block Client scripts as well.

Components would be the new place where logic is broken down from Page Scripts. Instead of scattering logic across many blocks, logic would be organized at the Component level. Components now have a single data script for dynamic data and a single client script to handle reactivity. Components can accept props, and the values of those props determine how dynamic values are resolved.

Assign dynamic data using the `component` dict (like `data` for page data scripts).
You can pass data from data script to client script using `component.component_data`.

In the client script you can access the component root using `this` keyword.

You also have access to `props` in both data and client script.

Demo - A Profile Card:

Screenshot 2026-06-29 at 10 35 43 AM

> Props in Components

Screenshot 2026-06-29 at 10 28 10 AM

> Component Data Script

Screenshot 2026-06-29 at 10 28 18 AM

> Component Client Script (JS) to add reactivity to elements

Screenshot 2026-06-29 at 10 28 21 AM

> Component Client Script (CSS) to add some animations

https://github.com/user-attachments/assets/ef7cc48d-43b0-4d61-959d-fabe725016c1

Another Demo - A Weather Card:

https://github.com/user-attachments/assets/6e89af5c-53e9-42ad-8f11-32ed9120e444


frontend/src/stores/componentStore.ts
```diff
@@ -9,13 +9,24 @@ import { createDocumentResource, createResource, toast } from "frappe-ui";
 import { defineStore } from "pinia";
 import { markRaw } from "vue";
 
+export const COMPONENT_DATA_FRAGMENT_KEY = "__fragment__";
+
+export type ComponentDataStore = Record<string, Record<string, Record<string, any>>>;
+
+const componentDataRequestSeq = new Map<string, number>();
+
+function getComponentDataRequestKey(componentId: string, instanceKey: string) {
```
stravo1: can there be a better way to avoid race condition?

builder/builder/doctype/builder_client_script/builder_client_script.json
```diff
@@ -37,9 +37,13 @@
    "link_doctype": "Builder Page",
    "link_fieldname": "builder_script",
    "table_fieldname": "client_scripts"
+  },
```
stravo1, line 40: does it even make sense to use Builder Client Scripts? Restrictions include site wide unique names, also public files for each script (which can be turned off if needed).

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -598,17 +598,20 @@ def get_block_html(blocks: str | list) -> tuple[str, str, dict, bool]:
...
 		tag = build_tag(block, shared_state)
 		# Add global script to the top
 		tag.insert(0, shared_state["global_script_tag"])
 
 		html = wrap_html_with_context(str(tag), block_context)
+
 		cleanup_props_stack(props, shared_state["standard_props_stack"])
+
 		html_parts.append(html)
+		print("".join(html_parts))
```
stravo1: TODO: Remove before merging, for debugging purposes only

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1002,42 +1015,93 @@ def get_visibility_condition_key(block: dict, data_key: dict | None) -> str | No
...
 
-	# Add local script to call the function
-	local_script = state["soup"].new_tag("script")
+	local_script = state["soup"].new_tag("style")
 	local_script.string = (
-		f"(client_script_{script_unique_id}).call("
-		f"document.querySelector('[data-block-uid=\"{{{{ unique_hash }}}}\"]'), "
-		f"{{{{ props | to_safe_json }}}}"
-		f");"
+		f"[data-block-uid=\"{{{{ unique_hash }}}}\"] {{ {script['script']} }}"
```
stravo1: Add docs saying how native CSS nesting is used to scope the styles, to apply on the container root the user has to use something like:
```css
&.sample-root {
...
}
```
instead of:
```css
.sample-root {
...
}
```

frontend/src/pages/PageBuilder.vue
```diff
@@ -272,10 +274,14 @@ watch(
 declare global {
 	interface Window {
 		blockController: typeof blockController;
+		canvasStore: ReturnType<typeof useCanvasStore>;
+		componentStore: ReturnType<typeof useComponentStore>;
 	}
 }
 
 window.blockController = blockController;
+window.canvasStore = canvasStore;
```
stravo1: TODO: remove before merging

builder/api.py
```diff
@@ -486,3 +486,19 @@ def get_codemirror_completions():
 def reorder_client_scripts(script_order: list[str]):
 	for idx, script_name in enumerate(script_order, start=1):
 		frappe.db.set_value("Builder Page Client Script", script_name, "idx", idx)
+
+
+@frappe.whitelist()
+def get_component_data(component_name: str, props: dict | str | None = None, script: str | None = None) -> dict:
+	"""Execute a component's data script and return the resulting data.
```
stravo1: added script param, mention in the docs

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1010,42 +1032,143 @@ def get_visibility_condition_key(block: dict, data_key: dict | None) -> str | No
...
+}
+
+
+def get_component_var_initial_value(var_def: dict):
+	if "initialValue" in var_def and var_def["initialValue"] is not None:
+		return var_def["initialValue"]
+	return DEFAULT_COMPONENT_VAR_VALUES.get(var_def.get("type", "string"), "")
+
+
+def format_component_var_entry(name: str, var_def: dict) -> str:
```
stravo1: rename as is used in pages as well

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -669,7 +681,7 @@ def build_tag(
 	return tag
 
 
-def get_block_context(block: dict, props: dict) -> dict:
+def get_block_context(block: dict, props: dict, component_id: str | None) -> dict:
 	"""
```
stravo1, line 708: little weird that the componet_id is just passed down

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1010,42 +1032,143 @@ def get_visibility_condition_key(block: dict, data_key: dict | None) -> str | No
...
+
+
+def create_local_client_script_tag(
+	state: dict, script_id: str, script: dict, block_vars: dict | None = None
+) -> bs.Tag:
+	"""Create a per-block script or style tag that invokes the global client script."""
+	if script["type"] == "JavaScript":
+		script_tag = state["soup"].new_tag("script")
+		invocation = create_client_script_invocation(script_id, script)
+		if script.get("from") == "component":
```
stravo1: it will always come from components only so no need of the next two lines

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1010,42 +1032,143 @@ def get_visibility_condition_key(block: dict, data_key: dict | None) -> str | No
...
+		else:
+			script_tag.string = invocation
+		return script_tag
+
+	style_tag = state["soup"].new_tag("style")
+	style_tag.string = f'[data-block-uid="{{{{ unique_hash }}}}"] {{ {script["script"]} }}'
+	return style_tag
+
+
+REACTIVE_COMPONENT_VAR_TYPES = {"object", "array"}
```
stravo1: move inside the required function

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1010,42 +1032,143 @@ def get_visibility_condition_key(block: dict, data_key: dict | None) -> str | No
...
+def build_page_vars_script(page_vars: dict | None = None) -> str:
+	page_vars = page_vars or {}
+	entries = [format_component_var_entry(name, var_def) for name, var_def in page_vars.items()]
+	if not entries:
+		return ""
+	vars_body = ", ".join(entries)
+	return f"const vars = {{ {vars_body} }};"
+
+
+def block_uses_reactive_vars(block: dict) -> bool:
```
stravo1: too many short functions

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1069,8 +1192,14 @@ def append_child_with_context(parent: bs.Tag, child: bs.Tag, context: dict):
 	if context.get("visibility_key"):
 		parent.append(f"{{% if {context['visibility_key']} %}}")
 
+	if context.get("component_id"):
+		parent.append(f"{{% with component = get_component_data('{context['component_id']}', props) %}}")
```
stravo1: can have better perf is we attach the data script to the block when we extend block, this way we avoid another DB call for every component as we are doing it once anyways

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1156,23 +1287,51 @@ def wrap_html_with_context(html: str, context: dict) -> str:
...
-		return block
+		return block, None
 
+	component_id = block.get("extendedFromComponent")
 	component = frappe.get_cached_value(
 		"Builder Component",
-		block["extendedFromComponent"],
-		["block", "name"],
+		component_id,
+		["block", "name", "component_js", "component_css", "component_props", "component_vars"],
```
stravo1: get everything atp

frontend/package.json
```diff
@@ -5,7 +5,8 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "vite build",
+    "build": "vite build && yarn build:runtime",
+    "build:runtime": "vite build --config vite.runtime.config.mjs",
```
stravo1: better name?

frontend/src/block.ts
```diff
@@ -994,26 +992,15 @@ class Block implements BlockOptions {
...
 		let blockProps = {};
 		if (this.isExtendedFromComponent() && !Object.keys(this.props || {}).length) {
-			blockProps = this.referenceComponent?.getBlockProps() || {};
+			blockProps =
+				useComponentStore().getComponent(this.extendedFromComponent as string)?.component_props || {};
 		} else {
 			blockProps = this.props || {};
 		}
-		return blockProps;
+		return blockProps as BlockProps;
```
stravo1: type conversion needed?

frontend/src/components/BlockLayers.vue
```diff
@@ -64,18 +64,14 @@
 								},
 							]"
 							aria-hidden="true"
-							v-if="!Boolean(element.extendedFromComponent) && !showCodeIcon(element)" />
-						<span
-							class="lucide-blocks mr-1 h-3 w-3"
+							v-if="!Boolean(element.extendedFromComponent)" />
+						<BlocksIcon
```
stravo1: check if using updated frappe-ui icons

frontend/src/components/BlockPropertySections/CollectionOptionsSection.ts
```diff
@@ -1,18 +1,33 @@
...
+import useCanvasStore from "@/stores/canvasStore";
 import usePageStore from "@/stores/pageStore";
 import blockController from "@/utils/blockController";
+import componentController from "@/utils/componentController";
 import { getRepeaterScopedData } from "@/utils/helpers";
 import { computed, h } from "vue";
 
 const keyOptions = computed(() => {
 	const pageStore = usePageStore();
+	const {editingMode, fragmentData} = useCanvasStore();
```
stravo1: breaks reactivity, can't destructure like this

frontend/src/components/BlockPropertySections/CollectionOptionsSection.ts
```diff
@@ -38,11 +53,14 @@ const keyOptions = computed(() => {
 	}
 
 	processObject(pageDataCollectionObject, "", repeatablePageDataKeys);
+	processObject(componentDataCollectionObject, "", repeatableComponentDataKeys);
 
 	const isPropsBasedRepeater = isInsideRepeater && repeaterDataKeyComesFrom == "props";
 	const repeatableProps: string[] = [];
 
-	const propsOfComponentRoot = blockController.getComponentRootBlock()?.getBlockProps();
+	const propsOfComponentRoot = blockController
```
stravo1: the previous version was fine, we need not pass anything to getComponentRoot

frontend/src/components/ClientScriptManager.vue
```diff
@@ -0,0 +1,32 @@
+<template>
```
stravo1: why even need this file now?

frontend/src/components/Controls/CodeEditor.vue
```diff
@@ -49,7 +49,7 @@ const props = withDefaults(
 	defineProps<{
 		modelValue?: Object | String | Array<any>;
 		type?: "JSON" | "HTML" | "Python" | "JavaScript" | "CSS";
-		mode?: "block" | "page";
+		mode?: "block" | "page" | "component";
```
stravo1, line 52: remove block?

frontend/src/components/Controls/CodeMirror/CodeMirrorEditor.vue
```diff
@@ -122,6 +125,16 @@ const allBlockProps = computed(() => {
 	};
 });
 
+const allBlockVars = computed(() => {
+	if (props.mode === "component") {
```
stravo1: can take care of autocomplete later

frontend/src/components/PageScript.vue
```diff
@@ -81,22 +102,63 @@
...
+								type="JavaScript"
+								label="Component JavaScript"
+								mode="component"
+								height="60vh"
+								:readonly="builderStore.readOnlyMode"
+								:autofocus="true"
+								@save="(value: string) => saveComponentClientScript('component_js', value)"
+								:showSaveButton="true"
+								:show-line-numbers="true"></CodeEditor>
+							<CodeEditor
```
stravo1, line 110: can they both not be moved into one? will probably lead to unsaved changes -> switch? check for dirty then before changing modes

frontend/src/components/PropsPopoverContent.vue
```diff
@@ -296,10 +296,11 @@ function filterDataOptions(dataArray: string[], query: string) {
 }
 
 const getOptions = async (query: string) => {
-	return filterDataOptions(pageDataArray.value, query).map((opt) => ({
+	const pageOptions = filterDataOptions(pageDataArray.value, query).map((opt) => ({
```
stravo1: remove non std. props related code

frontend/src/components/VisibilityInput.vue
```diff
@@ -21,36 +23,48 @@ const props = defineProps<{
 }>();
 
 const pageStore = usePageStore();
+const { editingMode } = useCanvasStore();
```
stravo1: again, breaks reactivity, can't destructure

frontend/src/pages/PageBuilder.vue
```diff
@@ -272,10 +275,14 @@ watch(
...
+window.componentStore = componentStore;
```
stravo1: remove before merge

frontend/src/stores/componentStore.ts
```diff
@@ -9,13 +9,35 @@ import { createDocumentResource, createResource, toast } from "frappe-ui";
...
+	return `${componentId}::${instanceKey}`;
+}
+
+function parseJSONWithFallback<T>(value: T | string | undefined, fallback: T): T {
+	if (value === undefined || value === null || value === "") {
```
stravo1: should be in utils

frontend/src/utils/blockController.ts
```diff
@@ -345,13 +352,11 @@ const blockController = {
...
 		return block;
 	},
-	getBlockClientScript: () => {
-		return blockController.getFirstSelectedBlock()?.getBlockClientScript();
-	},
-	setBlockClientScript: (script: string) => {
-		blockController.getFirstSelectedBlock()?.setBlockClientScript(script);
-	},
 	getBlockProps: () => {
+		const canvasStore = useCanvasStore()
```
stravo1: not needed right? componentController mutates the block directly

frontend/src/utils/useCanvasUtils.ts
```diff
@@ -281,3 +281,7 @@ export function useCanvasUtils(
 		isDirty,
 	};
 }
+
+export function markCanvasDirty(dirty: boolean = true) {
```
stravo1: maybe can be done directly in the required file.

builder/builder/doctype/builder_client_script/builder_client_script.json
stravo1: moved from Builder Client Scripts to separate fields under Builder Components Doctype

builder/builder/doctype/builder_page/builder_page.py
stravo1: uses get_cached_value so can't use "*"

frontend/src/components/PageScript.vue
stravo1: tried, does not reduce too much code because logic needs to bed added to handle other aspects

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -265,13 +266,13 @@ def publish(self):
 		self.published = 1
 		self.published_at = now()
 		if self.draft_blocks:
-			# snapshot the content going live (component versions pinned) before it overwrites `blocks`
+			# snapshot the content going live; blocks already carry componentVersion pins
+			# from when each component was used in the page (pinned at drag-drop)
 			take_snapshot(
 				"Builder Page",
 				self.name,
 				fields=["draft_blocks", "page_data_script"],
```
stravo1, line 274: currently only blocks are snapshotted, we should snapshot every aspect of the page

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1089,8 +1170,16 @@ def append_child_with_context(parent: bs.Tag, child: bs.Tag, context: dict):
 	if context.get("visibility_key"):
 		parent.append(f"{{% if {context['visibility_key']} %}}")
 
+	if context.get("component_id"):
+		parent.append(
+			f"{{% with component = get_component_data('{context['component_id']}', props, '{context['component_data_script']}') %}}"
```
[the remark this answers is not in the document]
stravo1: Introduced in https://github.com/frappe/builder/pull/627/changes/c3aeae6f6637b1a2a8b62224019df6735b4a70c4

Reverting as we are doing a fetch nonetheless.

builder/builder/doctype/builder_page/builder_page.py
```diff
@@ -1032,40 +1051,101 @@ def get_visibility_condition_key(block: dict, data_key: dict | None) -> str | No
...
+		script_tag = state["soup"].new_tag("script")
+		invocation = (
+			f"(client_script_{script_id}).call("
+			f"document.querySelector('[data-block-uid=\"{{{{ unique_hash }}}}\"]'), "
+			f"{{{{ component.component_data | to_safe_json }}}}, "
+			f"{{{{ props | to_safe_json }}}}, "
+			f"variables"
+			f");"
+		)
+		if script.get("from") == "component":
```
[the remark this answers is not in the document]
stravo1: @greptileai this is intended, component.compont_data can be assigned via component data script, so this is fine.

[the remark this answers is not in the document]
stravo1: @greptileai re-review with this in mind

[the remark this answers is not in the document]
stravo1: @greptileai yes, can you re-evaluate the score now

frontend/src/components/TextBlock.vue
```diff
@@ -57,14 +57,18 @@ const props = withDefaults(
...
 		breakpoint?: string;
+		parentBlockUid?: string | null;
 	}>(),
 	{
 		preview: false,
 		data: () => ({}),
+		componentData: null,
 		defaultProps: null,
 		breakpoint: "desktop",
+		parentBlockUid: null,
```
surajshetty3416: Why is this required 🤔 ?

use `block.parentBlock`?

frontend/src/block.ts
```diff
@@ -1113,6 +1156,124 @@ function syncBlockWithComponent(
...
+		}
+	});
+	return map;
+};
+
+// Rebuild a component instance's children from the new version's structure, preserving user
+// overrides on matched children (matched by referenceBlockId via three-way diff against the old
+// version). Children with no counterpart in the new version are dropped — component subtrees
+// can't host detached blocks. Nested component instances fall back to resetWithComponent.
+function rebuildWithComponent(
```
surajshetty3416: Lets move these out of `block.ts`

surajshetty3416 (changes requested):

frontend/src/components/TextBlock.vue
stravo1: these are remnants of block script, UID is used to differentiate between different blocks inside a repeater. because a single block object is used for all repeater children, we cant attach the data to the block object as it is shared.

different blocks in a repeater could have different block data, so to fetch the dynamic values (block data), TextBlock (and BlockHTML) used this parentBlockUid to fetch the block data from the blockDataStore (which is removed now).

but same is applicable for components: a component which is a child of a repeater will have only one block object re-used to display all the children. however right now i am passing the component data directly and using componentController to manage component data while editing a component. so currently they are not used anywhere and i will remove them.

all this explanation was to keep a note why they were used and where to look for if they are needed again.

frontend/src/block.ts
stravo1: refactored and moved functions to their own files: https://github.com/frappe/builder/pull/627/commits/d982168a537849d23da620391178aa85433e5b97

are the files paths (putting them under utils) correct?

surajshetty3416 (approved):

PR #635, opened 2026-06-19
feat: Add Persona Survey page and related settings

surajshetty3416:
Before user starts using Builder, understand the persona of the user so that we can tailor Builder accordingly
Screenshot 2026-06-19 at 2 03 49 PM


surajshetty3416:
@mergifyio backport master

PR #637, opened 2026-06-22
Remove unused path from Tailwind config

gaurahari-dasa:
The removed path refers to a node_modules folder in the parent directory which is clearly wrong as the node_modules folder in the current directory is where frappe-ui package will be found installed.


gaurahari-dasa:
I get it. That parent path to frappe-ui is required when we execute `yarn dev` or `yarn build` from the parent (builder) directory. Hence, this PR is not really required.

PR #647, opened 2026-06-29
feat: Unified Components and Templates

stravo1:
## Summary

This PR unifies Builder Templates and Components under `Builder Component`.

Templates are now standard/first-party components (`is_standard = 1`), while user components remain non-standard. Existing page content requires no migration because templates previously inserted copied blocks.

## Changes

- Added catalog metadata to `Builder Component`:
  - Standard status
  - Label and category
  - Preview image
  - Preview width and height
  - Sort order
- Converted shipped block-template fixtures into standard component fixtures.
- Updated install/migration sync to import standard components.
- Templates tab now shows only standard components.
- Components tab excludes standard components and retains page/global filtering.
- Added category grouping to both tabs.
- Added grid and list views to both tabs using `ItemListRow`.
- Templates now use the normal component insertion flow with pinned versions.
- Template instances support component props, data scripts, JavaScript, CSS, reset, detach, and updates.
- Added a component details popover for editing:
  - Label
  - Category
  - Preview image
  - Preview width and height
- Creating a template now creates a standard `Builder Component`.
- Standard components are exported as fixtures when saved in developer mode.
- `Block Template` remains as a deprecated compatibility surface but is no longer used by active builder flows.

## Migration

No page-content migration is required. Previously inserted templates already exist as ordinary copied blocks.

Existing `Block Template` records are retained but no new records are created.

Demo:

https://github.com/user-attachments/assets/2c5d9450-8453-4868-8767-16a0295db480

Screenshot 2026-06-29 at 8 03 14 PM


stravo1:
We need Component Preview. Closing this as this we have no immediate plans of unifying components and templates

PR #648, opened 2026-06-29
feat: Component slots

stravo1:
## Summary

Adds component slots so saved components can keep their internal block order locked while still allowing page authors to inject editable content into declared slot areas.

This introduces a Vue/React-style slot model for Builder components:
- Component authors can mark eligible container blocks as named slots.
- Page instances can edit content inside those slots without unlocking the rest of the component structure.
- Slot content is treated as detached instance content, so it can be deleted/reordered independently from component-owned blocks.
- Empty filled slots intentionally stay empty; deleting filled slot content does not restore fallback automatically.

### Component slot authoring

- Added slot metadata to blocks: `slotName` and `slotFilled`.
- Added a block property section for declaring named component slots.
- Validates component slots server-side:
  - slot names must be unique
  - slots cannot be nested
  - slots cannot live inside repeaters
  - slots must be declared on supported container blocks

### Slot behavior in component instances

- Slot fallback children are materialized as detached instance blocks.
- Blocks inside slots can be edited, deleted, duplicated, wrapped, and reordered independently.
- Component-owned blocks outside slots remain protected from structure/order changes.
- Filled slots are preserved by slot name when components are updated or synced.
- Empty filled slots remain empty instead of restoring fallback content.
- Added a “Reset Slot to Fallback” action for slot instances.

### Block layers UX

- Slot blocks now use a slot-style icon prefix instead of a visible tag.
- Drag/drop restrictions now respect the distinction between component-owned blocks and detached slot content.

### Component sync/update behavior

- Component sync preserves filled slot content where the slot name still exists.
- Generic warning is shown before component updates/syncs because slot removal/renaming may discard detached slot content.

Demo - Popover Component with a slot named `body`:

https://github.com/user-attachments/assets/14ac4d5d-fc1d-4860-9fc7-9a94ecc0aaa7

Screenshot 2026-06-29 at 9 16 06 PM

> Assigning slots


stravo1:
Will work on it when there is a requirement.

PR #656, opened 2026-07-02
feat: Canvases in iframes

stravo1:
Render each responsive canvas breakpoint inside an isolated iframe.

This makes the editor canvas behave more like the published page by separating its DOM, styles, events, and client scripts from the Builder interface.

- Styles and scripts no longer interfere with the editor UI
- Each breakpoint has its own isolated document and runtime
- Component, and page client scripts can be emulated in the canvas
- Canvas interactions operate across iframe boundaries

> This does not guarantee safe script execution as scripts can still cross the iframe boundary and reach parent window, but we do achieve total DOM isolation.

https://github.com/user-attachments/assets/996beca0-1c12-4b4e-a03d-fa9ed565ac08

https://github.com/user-attachments/assets/3a020ef4-3506-4e7b-8754-7a7f1403dc3d

Screenshot 2026-07-02 at 11 52 50 PM


stravo1:
This implementation has a real cost. Any implementation that uses an iframe has the same cost.

An iframe also gives the most complete isolation available. It isolates the styles and the code.

The shadow DOM gives a middle ground. It isolates the styles. Manual JavaScript sandboxing (facades) isolates the code.

I close this for now. The work needs more attention first, to reduce the performance cost.

PR #659, opened 2026-07-06
fix: Update frappe-ui to v1.0.0-beta.21

surajshetty3416:
Updates the frappe-ui submodule from v1.0.0-beta.8 to v1.0.0-beta.21 (keeping the exact version pin in `frontend/package.json` in sync so yarn resolves the workspace copy), plus the adaptations the new version needs:

## Adaptations

- **Typography tokens**: ran frappe-ui's `migrate-tokens-v2` codemod — 20 renames across 16 files (`text-5xl` → `text-4xl`, `text-xl-semibold` → `text-lg-semibold`, etc.).
- **Popover**: `popoverClass` is a no-op in the new Popover. Replaced the `!mt-[20px]` spacing hack in `BuilderToolbar` with the `:offset` prop, and dropped the already-inert `!min-w-fit` / `appendTo` props from `ColorPicker` and `TextBlockBubbleMenu`. (The deprecated-but-supported `placement` prop and `#target`/`#body` slots are left as-is.)
- **Types**: the stricter resource typings (`data`/`promise` now nullable, Badge `theme` a union, Combobox options) surfaced 24 new `vue-tsc` errors — fixed with null guards and narrow casts. Also mapped `vue` in `tsconfig.json` paths to the hoisted copy so vue-tsc stops resolving frappe-ui's types against its nested vue and creating type-identity mismatches. Net result: 55 errors vs 58 on develop (no new errors, 3 pre-existing ones fixed incidentally).

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -38,7 +38,7 @@
     "@vueuse/core": "^10.2.1",
     "autoprefixer": "^10.4.2",
     "codemirror": "^6.0.2",
-    "frappe-ui": "1.0.0-beta.8",
+    "frappe-ui": "1.0.0-beta.21",
     "js-yaml": "^4.1.1",
     "opentype.js": "^1.3.4",
     "pinia": "^2.0.28",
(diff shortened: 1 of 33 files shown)


surajshetty3416:
@mergifyio backport master

PR #661, opened 2026-07-07
fix: prevent arrow keys clobbering text InlineInput fields

### LINKED ISSUE #660: bug: Arrow keys overwrite text input fields with numbers (Video URL, etc.)
###  Description 
Pressing the Up/Down arrow keys inside plain-text property fields overwrites
the field's value with a number (or `NaN`). Arrow keys should only step
values in numeric fields, not text fields.

### Steps to reproduce
1. Add a **Video** block to a page.
2. Select it and open the **Video Options** section in the right panel.
3. Click into the **Video URL** field.
4. Press the **Up** or **Down** arrow key.

### Expected
Arrow keys move the text cursor / do nothing. The URL is preserved.

### Actual
The field value is replaced with `NaNfree`, `1`, `2`… destroying the entered URL.
Video: https://www.loom.com/share/c86d13deda514810ac9d91305b450c05

### Affected Fields
Any plain-text `InlineInput` field, including:

- Video Options → **Video URL**
- Accessibility → **Aria Label**
- Options → **Class**
- Input Options → **Placeholder** etc
Not affected: numeric fields with units (position, grid sizing) and slider
fields (grid Columns/Rows) arrow stepping as intended. Image URL is
also unaffected since it uses a different input component.

### Cause
`InlineInput.handleKeyDown` (`frontend/src/components/Controls/InlineInput.vue`)
runs numeric increment/decrement on ArrowUp/ArrowDown for **every** field,
with no check for whether the field is numeric.

### Suggested Fix
Gate the stepping so it only applies to numeric inputs(those with
`unitOptions` (px/%) or `enableSlider` (grid columns/rows)). Text fields have
neither and should fall back to native cursor behaviour.

dhruv2x:
`InlineInput.handleKeyDown` was applying ArrowUp/ArrowDown increment/decrement logic to every input, causing plain text fields (Video URL, Aria Label, Class, Placeholder, etc.) to be replaced with numeric values or `NaN`.

The stepping logic is now gated to numeric inputs only fields that expose `unitOptions` (e.g. `px`, `%`, grid sizing) or `enableSlider` (grid rows/columns). For all other inputs, ArrowUp/ArrowDown now fall back to the browser's default cursor navigation behavior.

Before fix when we do arrow key up and down:

https://github.com/user-attachments/assets/7ab4ce5b-a410-4c02-b76b-16b4fd17f4c7

After fix :

https://github.com/user-attachments/assets/9f3618fd-ae6e-401c-aaea-c178e7de0d0b

Fixes #660

### DIFF
--- a/frontend/src/components/Controls/InlineInput.vue
+++ b/frontend/src/components/Controls/InlineInput.vue
@@ -139,6 +139,8 @@ const handleMouseDown = (e: MouseEvent) => {
 };
 
 const handleKeyDown = (e: KeyboardEvent) => {
+	// Arrow-key stepping only for numeric inputs, not plain text fields.
+	if (!props.unitOptions.length && !props.enableSlider) return;
 	if (e.key === "ArrowUp" || e.key === "ArrowDown") {
 		const step = e.key === "ArrowUp" ? 1 : -1;
 		incrementOrDecrement(step);


dhruv2x:
not sure who to tag, but @surajshetty3416, could you have a look please?

surajshetty3416:
@dhruv2x can you add GIF demonstrating the fix?

dhruv2x:
> can you add GIF demonstrating the fix?

I've added in PR description

surajshetty3416 (approved):

PR #671, opened 2026-07-14
feat: Bob - AI Assistant for Builder  🤖

surajshetty3416:
Bob is an AI assistant inside the Builder editor. One prompt builds a page. The same conversation then keeps editing it, and writes the design tokens, page scripts, DocTypes and form wiring that go with it.

It runs as one loop with 28 tools. There is no "generate mode" and "edit mode": what changes between turns is which tools the model reaches for.

## Prompt to a working site

A fine jewellery house in Paris, from a single prompt. Bob searches for photography and asks which hero shot to open on, mints its own design system, then writes the page while the canvas follows it down: 94 blocks, a full-bleed hero, the collection, the atelier, an enquiry form.

Then it asks for the one thing it cannot decide alone, whether to save that form's submissions to a new DocType. Approve, and it creates the DocType, wires the form and hands back a link. The rest of the video is that promise being checked: publish, send a real enquiry from the live page, then follow Bob's own link into Desk and find the record sitting there.

https://github.com/user-attachments/assets/f7955dab-4342-4af9-bf45-a13a19c8dd7d

## Then keep editing it

Select a block on the canvas and it joins the conversation as context. Three asks, three different paths through the same loop: the headline is rewritten in place, one block, four seconds; "deepen the gold, warm the ivory" is answered by editing two design tokens rather than the page, so the whole site retones at once; and "a slow fade-up as each section scrolls in" lands as two page scripts, the CSS and the JavaScript kept separate. Publish, and the motion runs for real.

https://github.com/user-attachments/assets/01649d8b-f773-4092-9cc9-575b86276617

## In the editor

Bob lives in the left panel. Every turn leaves a trail you can open: what it thought, what it ran, what it changed, how long it took, and a Revert for that turn alone.

It asks everything it needs in one card instead of a wall of questions, and the card collapses into the answers once it is filled in.

A form it wired, with the enquiry the live page just sent, in Desk.

## What it can do

- **Build** a full page from a prompt, streamed into the canvas as it is written, against a design system it mints first
- **Edit** surgically or page-wide, with the blocks you select on the canvas as context
- **Style** by token, so a retheme is a token edit rather than a rebuild
- **Script** a page, CSS and JavaScript kept as separate page scripts
- **Wire data**: inspect the site's DocTypes, propose new ones, seed sample rows, write the page's data script, and connect a form so submissions actually save
- **Look at its own work**: screenshot the draft and review it before finishing
- **Remember** durable facts about a site across chats

Anything sensitive (a new DocType, sample data, the home page, global code, form wiring) stops and asks first.

## Configuration

Settings → AI walks through connecting a provider. Presets ship for OpenRouter, Anthropic, OpenAI, ChatGPT (through a Codex subscription rather than a key) and Google Gemini, plus a custom option for any OpenAI-compatible gateway (OpenCode, Ollama, vLLM, a private proxy), so a low cost coding subscription can drive the loop.

Each preset switches on the models worth starting with: OpenRouter opens with Claude Sonnet 5 and GPT-5.6 Luna, and leaves GPT-5.6 Terra, Gemini 3.6 Flash and a free Nemotron one tick away. A provider's full catalog can be imported from its `/v1/models` endpoint, and context window and vision support are filled in on save so adding a model only asks for its id.

---

<details>
<summary><b>How it works</b> — architecture, the tool inventory, how the model is steered, testing, known gaps</summary>

### The server is authoritative for every turn

The page is loaded from the database into a mutating `WorkingTree` (`builder/ai/agent/tree.py`). Every block op is applied there first, so the tool result handed back to the model is honest ("applied to block X", "3 of 12 not found"), and a wrong ref drives a self correcting round instead of a silent divergence. Ops the tree accepts are then mirrored to the canvas, which is a live view rather than the source of truth. Ops it rejects are never emitted, so the two sides cannot drift.

### One loop, one registry

`AgentRunner` (`builder/ai/agent/loop.py`) holds per turn state, builds the message list and drives the tool calling loop. Tool behaviour lives in the registry (`builder/ai/agent/registry.py`); the loop never changes when a capability is added. Every tool declares a *side*:

| Side | Meaning |
|---|---|
| `client` | a page edit: applied to the WorkingTree, then mirrored to the canvas |
| `server` | the loop runs the handler and feeds its string result back to the model |
| `terminal` | the call ends the turn and hands control back to the user (a question, a plan, a confirm gate) |

A tool may also produce a large *streamed artifact*. `generate_page` sets `artifact="page_yaml"`, so the loop hands execution to a generator that streams the YAML to the client for a live preview and returns the canonical op to apply. The agent calling the tool is the signal to generate: no out of band state decides it.

### Generation path

YAML is expanded to the editor's serialized block shape and written to `draft_blocks` server side (`builder/ai/page_writer.py`). The frontend keeps a small port of the conversion purely for the streaming preview; the authoritative tree always comes from the server, with the same block ids.

### Transactions

A turn runs as a background job, and a job that raises rolls its whole transaction back, so work has to be made durable as it goes. `AgentRunner.checkpoint()` commits once at the end of each round. Tools never commit. Any event whose handler re-reads what the turn just wrote is emitted with `after_commit=True`, so frappe holds it until the transaction lands and drops it on rollback: the client and the database cannot disagree.

### Concurrency

Run locks are atomic Redis `SET NX EX` with a fencing token (`builder/ai/locks.py`), so a crashed worker cannot wedge a page or session: the TTL expires, and a worker that outlived its TTL cannot release a newer holder's lock. This is generic enough to upstream as a `frappe.cache().lock(...)` primitive, with Builder as the first consumer.

### Realtime contract

Eight events over Redis pub/sub, each suffixed with a channel (the page id in the editor, the session id for a page-less chat): `ai_chat_step`, `ai_chat_stream`, `ai_chat_progress`, `ai_chat_tool_batch`, `ai_chat_clarify`, `ai_chat_complete`, `ai_chat_error`, and `ai_chat_refetch` — the one a server tool fires when it changes state the canvas only loads at editor start (theme variables, page data, the page doc), so the client refetches just those. Live narration emits immediately; anything the client reads back from the database is deferred to the commit. `ai_chat_step` carries the whole turn timeline, a thought, a tool call or a chunk of text, each with its own status and duration, which is what the chat renders as the collapsible trail.

### Revert

Pre turn snapshots reuse Builder's generic Builder Snapshot system, tagged `snapshot_type="AI"` (`builder/ai/snapshots.py`). State is captured synchronously at request time, before the worker or any streaming touches the canvas, and the snapshot doc is created only when a turn actually mutates a page, so clarify and plan turns leave nothing behind. One revert restores blocks, the data script and the page's client scripts together.

### Token efficiency

`BlockCodec` (`builder/ai/block_codec.py`) compresses blocks into a compact YAML vocabulary for the prompt and expands them back. Large pages are sent as an outline (one line per block) with full detail only for the blocks the user has selected, so a targeted edit costs no read round trip. Prompt caching uses pinned breakpoints on the system prompt plus rolling per turn markers.

### Data model

Builder AI Session, Builder AI Message, Builder AI Memory, Builder AI Provider, Builder AI Model.

### The 28 tools

**Build** — `generate_page`, `set_design_token`, `extract_component`
**Edit** — `update_block`, `update_blocks`, `add_block`, `remove_block`, `move_block`, `read_block`, `query_blocks`
**Scripts** — `set_page_script`, `update_script`, `get_page_scripts`
**Data** — `list_doctypes`, `get_doctype_schema`, `query_records`, `get_document`, `create_doctype`, `seed_sample_data`, `write_page_data_script`, `connect_form`
**Page and site** — `set_page_settings`, `set_home_page`, `edit_global_settings`, `preview_page`
**Conversation** — `present_ui`, `remember`
**Assets** — `search_images`

### Steering the model

Most of the work here is not plumbing, it is making a general model behave like a designer who knows Builder.

- **Role specific system prompts**: an editor agent prompt, a generation prompt written as an art direction brief, and a component prompt.
- **Structural validation with teaching errors.** The working tree rejects the mistakes models actually make and explains the fix in the tool result: expressions used as binding keys, repeaters "created" through `bind`, Jinja or moustache text in content and attributes (which crashes the published route), and CSS injected from JavaScript instead of a CSS script.
- **Corrective rounds.** A no-tool round that should have been a tool call gets exactly one correction, across three shapes: a card written as plain text, options asked as prose bullets, and a summary claiming an edit that was never applied. A build that minted a design system but never called `generate_page` has its own one-shot correction on a separate budget, because that turn did mutate the page.
- **Confirm gates.** Anything sensitive ends the turn and asks. The privileged write runs in the user's own request, never inside the model's turn, and the outcome is written back into the conversation so the next turn knows what happened.
- **Self review.** `preview_page` screenshots the draft and asks the model to review its own output for breakage and template fingerprints, on vision capable models only. The capture is tall and handed over as readable tiles, because a single full-page screenshot arrives at the model too small to read.
- **Ordering rules.** Scripts are written after the page exists, so selectors target classes the model can see rather than ones it is guessing at.

### Also in this PR

- Parallel chat sessions per page, with a session switcher and history
- Context window usage per turn, and per-call token counts in the debug panel
- Cancel mid turn: the stream closes, so providers stop billing further tokens
- Tolerant parsing for weaker models that emit tool calls as plain text JSON
- A debug panel for the raw turn signals

### Testing

- `builder/ai/test_tree_apply.py`: 16 tests covering the mutating tree, uniform and per block updates, and the rejection rules
- `builder/ai/test_page_writer.py`: 10 tests covering YAML expansion, style normalization and the repeater data shim
- Manual verification against a live site for the flows that need a real model

### Known gaps

- Coverage on `block_codec.py` and `selectors.py` is thin
- CSS page scripts are applied on the published page but not inside the editor canvas, so a page whose styling lives in a CSS script renders differently in the two places
- Builder AI Batch and Builder AI Batch Task doctypes are present but unused: the parallel fan-out that would drive them (a dashboard level orchestrator spawning one sub-agent per page) is not part of this PR

</details>

<details>
<summary><b>How it got here</b> — five months of experiments, dead ends and refinements</summary>

365 commits between 2 March and 11 August, plus the merges from develop along the way. The shape changed thoroughly along the way, and most of what was learned is not visible in the final diff.

### March: a modal that generated a page

The first cut was not a chat. It was a button in the toolbar that opened `AIPageGeneratorModal`, took a prompt, and streamed a page in. The interesting problems were all about the stream: models emit broken JSON halfway through, so there was a JSON repair pass; the canvas thrashed while blocks arrived, so streaming got throttled; page saving had to be paused while blocks were still landing.

Two decisions from that month survived everything else. **Blocks are described in YAML, not JSON** ("Update prompts and parsing to use YAML format for block structures to save credits") because the block tree is the expensive part of every prompt. And the model is given **an art direction brief rather than a schema**, because a schema produces pages that look like a schema.

The month ended with the first edit affordances: rewrite this text, replace this image, a keyboard shortcut to open the modal on a block, and a preset picker to make prompting easier.

### April: the chat panel

Summaries in markdown, selected blocks and attached images as chips, and the first tools that could fetch and update page scripts on demand. Generation was still one thing and editing another.

### May: one loop

"Use single agentic loop for all flows" is the pivot the whole PR rests on. Before it there were separate paths for generating, modifying a block, rewriting text and replacing an image, each with its own prompt and its own parsing. After it there is one tool calling loop, and generation is just the tool a model reaches for when the page is empty. Everything since has been adding tools rather than adding modes.

### June: the server takes over

Multi-round edits went live with per-turn revert on top of Builder Snapshot, and client ops started being validated server side with corrections logged. This is where the `WorkingTree` came from: the client had been the source of truth, the model would name a block that did not exist, and the two sides would quietly disagree. Telemetry for tokens and timing landed the same month, which is how the cost line in the chat exists.

### July: the data layer, gated

DocTypes, sample data, form wiring and global settings. All of it is capable of doing real damage, so all of it stops and asks, and the privileged write happens in the user's own request rather than inside the model's turn. Confirm cards took a few passes to become readable (no duplicate title, no raw JSON dump). Persistent memory arrived at the end of the month.

### August: configuration, and the last mile

Providers and models became doctypes instead of a hardcoded list, so a site can add any OpenAI-compatible gateway. A guided setup flow replaced "paste a key somewhere". The turn timeline started streaming. Questions collapsed into a single card. A ChatGPT-subscription provider went in via Codex OAuth.

### Tried and dropped

- **Model tiers with automatic task classification.** The agent used to classify the task and pick a cheap or expensive model for you. It was replaced by one model you choose. Guessing wrong was worse than the money it saved.
- **Sending the page as an outline.** Written to save tokens, then widened twice ("give the agent the whole page, not an outline of it") once real pages turned out to fit. The outline survives only for genuinely large pages.
- **Tokenising spacing and sizes.** Design tokens now cover colours, fonts and dimensions. Spacing tokens made every edit a two step negotiation.
- **Per-turn cost in the site's currency.** Shipped in July, dropped in August with the default-model flag. Pricing drifts per provider and the number was more often wrong than useful; token counts and context usage stayed.
- **Stock video search**, removed.
- **Synchronous stream sync**, reverted a month in for unreliability.
- **The generator modal itself**, deleted once the loop could do everything it did.

### Refinements that mattered more than they look

- Ask everything in one card, not one question per turn
- Name each token in the timeline, and collapse repeated rows
- Say what Bob did in the user's words, not the tool's
- Search photos per subject, not once per page
- A tool that raises no longer kills the turn
- A mistyped image URL no longer kills the build
- Add a subtree in one call, not one block per round
- Show the model the whole page, not its top third

### The other branches

Plenty was tried outside this branch and is parked rather than merged. Mostly prefixed `ai-`, all pointed at the same question from different angles:

| Branch | What it was trying |
|---|---|
| `ai-driven-page-builder` | the original generator line, where the March work started |
| `ai-selector-bulk-edit`, `ai-bulk-edit-2` | editing by CSS selector, so one instruction could restyle every matching block |
| `ai-sessions-all-in` | parallel chat sessions before they became part of this branch |
| `ai-first-builder` | an AI-first entry point: start from the dashboard and watch a page assemble, no editor first |
| `ai-codebase-context` | a primitives-only agent — `run_python` and source access instead of block tools — plus creative option cards and font specimens |
| `ai-unified` | the merge of those two: multi-page sites, site consistency between pages, a build pill that survives navigation |
| `agent-v2` | a from-scratch rewrite of the loop with the transcript on the session doctype and componentization built in |
| `ai-design-agency` | the agency journey and a studio mode |
| `ai-api-base-override` | routing completions through any OpenAI-compatible `ai_api_base`, which is where the plain-text tool call salvage came from |

Some of it will come back. `ai-first-builder`'s dashboard chat and `ai-codebase-context`'s option cards are the two most likely.

### The story

The first version was a slot machine. You typed a sentence, pulled the lever, and a whole page fell out. When it was good it felt like magic. When it was 90% good, which was most of the time, your options were to accept the 10% you disliked or pull the lever again and lose the 90% you liked. People pulled the lever a lot.

So we added a second button for editing a block, and a third for rewriting text, and a fourth for replacing an image. Each one worked. Each one had its own prompt, its own parser, its own bugs, and none of them knew what the others had done. The model would rewrite a headline with no idea a design system existed.

The fix was to stop adding buttons. One conversation, one loop, and a set of tools the model picks from. Generation stopped being a mode and became a tool that gets used when the page is empty. That is the whole idea, and everything after it was consequences: if the model can edit the page, the server has to know what the page actually is; if it can create a DocType, it has to ask first; if a turn can go wrong, one click has to undo it.

The last thing to change was the smallest. For months it answered questions one at a time, like a form that arrived a field per day. Now it asks everything at once, in one card, and gets out of the way.

</details>


federicocalvo:
Hi @surajshetty3416 I don't know if you've considered making the AI options more configurable.

1- Being able to choose the models. Since they're hardcoded, you lose the option to try others, or to use new ones that are less expensive or more optimal until an update is released. There could obviously be recommended models.

2- Make it more open, like Flow, with more freedom to connect OpenRouter, OpenIA, Ollama, etc.

3- This is a bit of a long shot, but the option to connect it to a CLI (for those who use it locally to set up a website, etc.)

surajshetty3416:
@greptile-apps review

PR #673, opened 2026-07-15
chore: For v1.30 Release

surajshetty3416:
Release PR for v1.30, merging `develop` into `master`.

## Drag to reorder blocks on the canvas

Rearranging the layout no longer means dragging items in the layers tree. You can now grab any block directly on the canvas and drop it into a new position. The engine is pointer-based and zero-reflow, so blocks glide into place without jitter or layout shift, and it understands flex and grid containers: it shows where a block will land, handles nesting into and out of containers, and measures the breakpoint you are actually editing.

https://github.com/user-attachments/assets/3e88e40f-3d82-41b2-83f6-268e4c334fe0

https://github.com/user-attachments/assets/13b8693d-9e59-440e-be6c-fa4e959bfae1

## Rotate blocks and unified transform controls

Blocks can now be rotated right on the canvas. Four corner grips let you spin a block about its center, with Shift snapping to 15 degree steps, and a rotate control is exposed in the Style section. The whole editing chrome is rotation-aware: the selection box, resize handles, and the padding and margin handlers all stay aligned to a rotated block, and their cursors turn to match the on-screen angle. Corner resize is available on all four corners, and anchors are preserved when resizing a rotated block.

https://github.com/user-attachments/assets/35aa1958-e883-4238-ac2c-f41e80bdd0fa

## Inline template previews with device breakpoints

Previewing a template no longer opens a new tab. It now renders inside the template picker as an embedded, live preview of the page, with a desktop, tablet, and mobile toggle so you can check how a template responds before you use it. Pages render at true viewport widths and scale consistently so type stays proportionate across breakpoints. The picker also gains category filters, so you can narrow templates down to what you are building.

## Sharper inline text and SVG rendering

Inline text blocks such as span and anchor pills now render on the canvas exactly as they do on the published page. Previously the canvas wrapper fragmented the inline box, so pill backgrounds shattered into detached caps and text could go invisible on same-tone sections. Editing text no longer wraps content in a stray paragraph, which had corrupted saved markup. Inline SVG icons now render as HTML rather than editable text, fixing a canvas crash and a recursive update loop on large pages.

## Enhancements and fixes

- Refreshed the UI to frappe-ui v1.0.0-beta.21.
- Arrow keys no longer clobber text while editing InlineInput fields.
- Published page routes are proxied to Frappe in the Vite dev server, so previews work in local development.
- Client scripts are stripped from the page block JSON to keep saved pages clean.
- Clearer confirmation wording when copying an entire page in the clipboard flow.
- Consistent root background color across light and dark themes.


surajshetty3416:
Superseded by #674, which merges the same develop content into master via the conflict-free `for-release-v1.30.0` branch (resolves the TemplateGallery.vue conflict by keeping develop's version, which already integrates both the category filters and the inline template preview).

PR #682, opened 2026-07-20
feat: load real italic faces through the ital axis

surajshetty3416:
## Problem

The Google Fonts loader on published pages only carries the `wght` axis. Any italic text (`fontStyle: italic` on a block, or `<i>`/`<em>` inside `innerHTML`) renders as a browser-synthesised slant instead of the family's real italic faces. Template authors currently work around it with a hand-written `@import ...:ital@...` CSS client script per template.

Synthesised shear (top) vs the real italic faces this PR loads (bottom), Bodoni Moda:

<details>
<summary>Full before/after (Fraunces + Bodoni Moda)</summary>

</details>

## What this does

Italic usage is resolved on the block tree with CSS cascade semantics, in `build_tag`, right where the font-family cascade is already resolved: a block's own `fontStyle` wins, otherwise italic is inherited from the ancestor exactly like `ancestor_font`. That covers:

- `fontStyle: italic`/`oblique` on the block itself (any device style),
- a child that sets `fontStyle` without a family → italics land on the inherited font,
- an italic parent whose child only switches `font-family` → the child still renders italic (font-style inherits in CSS), so the child family gets its italic faces too; a child resetting `fontStyle: normal` breaks the cascade again,
- `<i>`/`<em>` tags and inline `font-style: italic` spans inside `innerHTML`.

`set_fonts` itself is untouched. Italic faces are folded into the **same single stylesheet request** per family via the `ital` axis:

```
before  https://fonts.googleapis.com/css2?family=Fraunces:wght@350;400;500&display=swap
after   https://fonts.googleapis.com/css2?family=Fraunces:ital,wght@0,350;0,400;0,500;1,300;1,400&display=swap
```

No font catalog is consulted for this: the css2 API silently drops `ital`/`wght` tuples a family doesn't ship (upright faces still load and the browser falls back to synthetic italics, i.e. today's behaviour), so the `ital` axis is safe to request unconditionally. 400 italic always rides along, so a family missing the exact italic instance still serves its regular italic face. Families used only upright keep byte-identical URLs to today.

### DIFF
--- a/builder/builder/doctype/builder_page/builder_page.py
+++ b/builder/builder/doctype/builder_page/builder_page.py
@@ -666,7 +666,11 @@ def get_block_html(blocks: str | list) -> tuple[str, str, dict, bool]:
 
 
 def build_tag(
-	block: dict, state: dict, data_key: dict | None = None, ancestor_font: str | None = None
+	block: dict,
+	state: dict,
+	data_key: dict | None = None,
+	ancestor_font: str | None = None,
+	ancestor_italic: bool = False,
 ) -> bs.Tag:
 	"""
 	Transforms a single block to an HTML tag.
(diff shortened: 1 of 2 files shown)


surajshetty3416:
Dropped the `fontList.json` dependency (7214d9eb): it's unmaintained, and it turns out the css2 API doesn't need guarding at all. It silently drops `ital`/`wght` tuples a family doesn't ship instead of rejecting the request (verified against static families like PT Serif, italic-less families like Oswald and Anton, and variable-italic families like Lora). So the `ital` axis is now emitted unconditionally whenever italics are used, with 400 italic always included as a fallback instance. No catalog, no network calls, and upright-only families keep byte-identical URLs.

PR #684, opened 2026-07-20
feat: preview state styles during canvas drags

stravo1:
When a state variant is active, canvas drag handles now read and write the active style key instead of falling back to the base style. This keeps rotation, resize, spacing, movement and border radius previews aligned with the selected state.

https://github.com/user-attachments/assets/44cb1171-5103-449e-b7d0-0a0f8cbf43a3


surajshetty3416:
@mergify update

PR #685, opened 2026-07-21
[WIP] feat: raw styles now support breakpoints

stravo1:
Previously Raw Styles were applied across all breakpoints, now Raw Styles are also breakpoint specific, which means users can apply different Raw Styles to blocks in different breakpoints.


stravo1:
Moved to this: https://github.com/frappe/builder/pull/691

PR #691, opened 2026-07-23
feat: replace raw styles with curated base styles editor

stravo1:
## Summary

Replaces the free-form Raw Styles panel with a **More Styles** panel backed by a curated CSS property list, so custom styles are validated against known CSS properties and merge directly into `baseStyles`, `tabletStyles` and `mobileStyles` instead of living in a separate `rawStyles` bag. This also means that the restriction of `rawStyles` being breakpoint independent is no more, users can set various custom styles in different breakpoints. 

- New `MoreStylesPanel` lets users add any CSS property from a searchable (fuzzy-matched), autogenerated property list, with support for state variants (hover/active/focus) and responsive breakpoints, consistent with the rest of the style editor
- CSS property metadata is generated by a script (`generate-css-property-metadata.mjs`) instead of hand-maintained
- Legacy `rawStyles` on existing pages, components, and block templates are normalized into `baseStyles` at render time, so previously published pages keep rendering correctly without a blocking migration

https://github.com/user-attachments/assets/52cf2a37-8d21-4ae9-b5e5-e2bc5e27f50c


surajshetty3416:
@stravo1 will this be completely backward compatible? Does copy paste work from old site with "rawStyles"?

stravo1:
> @stravo1 will this be completely backward compatible? Does copy paste work from old site with "rawStyles"?

Yes, it is!

builder/builder/patches/migrate_raw_styles_to_base_styles.py
```diff
@@ -0,0 +1,33 @@
+import frappe
```
surajshetty3416: Patches are scary in general... we can skip it in this case I think.

Let's remove this migration script? Editor should ideally auto heal the data.

surajshetty3416 (changes requested):

builder/builder/patches/migrate_raw_styles_to_base_styles.py
stravo1: yes i was thinking of removing patches, editor does auto-heal. thanks for confirming

frontend/src/utils/curatedStyleProperties.ts
```diff
@@ -0,0 +1,75 @@
+import accessibilitySection from "@/components/BlockPropertySections/AccessibilitySection";
```
surajshetty3416: Can't this be done via "BlockProperties.vue" instead of importing sections here again? We can just iterate over [sections](https://github.com/frappe/builder/pull/691/changes#diff-ebdcedcfd5c0398c148a7259bc1ca58a6a0667c061ce4d08339708861130212aR111)

Also, not really liking the naming for this. 

could simply be `usedStyleProperties` in section instead of `ownedStyleProperties` and `stylePropertiesWithControls` instead of `curatedStyleProperties`? 

> Let's try to use terminologies that we already have instead of introducing new ones

---

BTW, good idea to maintain style property list in controls/sections.

surajshetty3416 (changes requested):

surajshetty3416 (approved):

PR #694, opened 2026-07-24
fix: resolve text block bubble menu color picker positioning, slider drag, and popup layout

SmrutiSawant:
### Summary of Changes
Fixes positioning and interaction issues with the text editor's `TextBlockBubbleMenu` color picker:
- **Position Stability**: Anchors bubble menu coordinates relative to the text editor element 
- **Slider Dragging**:  Hue and Alpha sliders can be dragged smoothly without being blocked by parent toolbar event handlers.

Before:

https://github.com/user-attachments/assets/a9375364-98d2-4048-88db-c5f88982910d

After:

https://github.com/user-attachments/assets/ab491e96-f2b7-45c7-acbb-7b49c59e9c2d

### DIFF
--- a/frontend/src/components/Controls/ColorInput.vue
+++ b/frontend/src/components/Controls/ColorInput.vue
@@ -4,6 +4,7 @@
 			ref="colorPickerRef"
 			:placement="placement"
 			:offset="popoverOffset"
+			:portal-to="portalTo"
 			@open="events.onFocus"
 			@close="handleClose"
 			:modelValue="resolvedColor"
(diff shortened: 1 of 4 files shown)


stravo1:
The main issue was that the `TextBlockBubbleMenu` was hidden after losing focus when the user clicked the `ColorPicker`. A more concise fix was to keep it open while the `ColorPicker` remained focused.

The `Autocomplete` for the `ColorPicker` also jumped to the corner of the screen. Both issues have been fixed in: https://github.com/frappe/builder/pull/695

PR #697, opened 2026-07-28
fix: keep newly created text block editable through the creating click

stravo1:
Selecting synchronously in makeBlockEditable prevents the block's own deferred selectBlock call from clearing editableBlock right after it's set. TextBlock also swallows the one spurious click-outside event that the creating click's native "click" triggers before the block has rendered, and catches up the focus/edit-mode entry that the isEditable watch missed because the editor didn't exist yet at mount time.

Before:

https://github.com/user-attachments/assets/a3c125ce-50ba-465e-b896-3156d0652a92

After:

https://github.com/user-attachments/assets/49c85344-9456-4bb8-88ca-03dedf7f8c38


surajshetty3416:
We'll come up with a better solution

PR #699, opened 2026-07-28
feat: Add GapControl component with axis split mode

SmrutiSawant:
**Description :** 
Adds a new GapControl.vue component that supports both uniform mode (gap) and split mode (row-gap and column-gap) for Flex and Grid layout panels.

Before : 

https://github.com/user-attachments/assets/08830e31-22d3-46a2-8d69-08d4e4662c80

After : 

https://github.com/user-attachments/assets/70e58056-b8fd-4470-9e88-721b567ea838


frontend/src/utils/cssUtils.ts
```diff
@@ -344,6 +344,30 @@ function collapseBoxShorthand(parts: unknown[]): string {
 	return [top, right, bottom, left].join(" ");
 }
 
+/**
+ * Expands a CSS gap shorthand into its two component values (row-gap, column-gap).
+ * @param value - Gap shorthand value string
+ * @param fallback - Value used when the shorthand is empty
+ * @returns Array of exactly two values: [row-gap, column-gap]
+ */
+function expandGapShorthand(value: unknown, fallback = "0"): [string, string] {
```
stravo1: instead of adding shorthand for separate properties can we have a function like:

```
type ShorthandSlots = 2 | 4;

expandShorthand(given, slots) {...}

expandBoxShorthand(value) { expandShorthand(value, 4) }
expandGapShorthand(value) { expandShorthand(value, 2) }

```

we can also remove `fallback`? unnecessarily puts 0 when split from `unset`

frontend/src/components/GapControl.vue
```diff
@@ -0,0 +1,115 @@
...
+const normalize = (value: BoxValue) => normalizeValueWithUnits(String(value || "0"), "px");
+
+const toModelValue = (parts: BoxValue[]) => {
+	const [first, second] = parts;
+	return isColumnDirection.value
+		? collapseGapShorthand([second, first])
+		: collapseGapShorthand([first, second]);
+};
+
+const getMergedValue = (parts: BoxValue[]) => parts[0] ?? 0;
```
stravo1: some code for BorderRadiusControl, SpacingControl and GapControl are repeating, maybe we can have them as defaults in SplitInputMode like `getControlAttrs`, `getMergedValue`, etc.

stravo1 (changes requested):

frontend/src/components/GapControl.vue
```diff
@@ -0,0 +1,104 @@
...
+import { computed } from "vue";
+
+const isColumnDirection = computed(() =>
+	String(blockController.getNativeStyle("flexDirection") || blockController.getCascadingStyle("flexDirection") || "").startsWith("column"),
+);
+
+const splits = computed(() => isColumnDirection.value ? [{ label: "H" }, { label: "V" }] : [{ label: "V" }, { label: "H" }]);
+
+const getBlockGap = (block: Block): [string, string] => {
+	const [exR, exC] = expandGapShorthand(block.getNativeStyle("gap"));
```
stravo1: we prefer longer, more readable variable names over shorthands/abbreviations, so instead of `r` and `c`, we can use `row` and `column`.

you can check this document: https://github.com/frappe/pilot/blob/develop/CLAUDE.md#code-taste

stravo1 (changes requested):

frontend/src/components/GapControl.vue
stravo1, line 1: we can delete this file as well like I have done with BorderRadiusControl, BorderControl, SpacingControl.

take a look at SpacingSection, the reason why that code is so concise is because it hides the complexity in blockController.getMargin(...) and blockController.setMargin(...) (or padding)

almost same pattern is repeated here, so we can follow the codebase and put the code in blockController.getGap(...) and setGap(...) ?

stravo1 (changes requested):

frontend/src/components/BlockPropertySections/LayoutSection.ts
```diff
@@ -2,8 +2,29 @@ import BlockFlexLayoutHandler from "@/components/BlockFlexLayoutHandler.vue";
 import BlockGridLayoutHandler from "@/components/BlockGridLayoutHandler.vue";
 import OptionToggle from "@/components/Controls/OptionToggle.vue";
 import blockController from "@/utils/blockController";
+import { collapseGapShorthand, expandGapShorthand } from "@/utils/cssUtils";
 import StylePropertyControl from "../Controls/StylePropertyControl.vue";
 
+const SPLITS = ["V", "H"];
```
stravo1, line 8: The gap shorthand is always row-gap then column-gap. CSS does not
reinterpret it by flex-direction, so both split orders wrote the same
value and the index swap only undid the label flip.

stravo1 (approved):

PR #700, opened 2026-07-28
fix: keep autocomplete lists full while filtering

stravo1:
Each `getOptions` implementation filtered and capped its own list. Once an item was selected the list showed no extra options.

Add `filterOptions`, which puts the matches for a query first, fills the rest of the list with the other options, and stops at the limit. This way some options other than the selected one stay visible when available.

`filterOptions` takes `limit` and `windowRadius` in one options object. Without `windowRadius` it keeps the matches first, then fills the rest of the list. With `windowRadius` it keeps each match at its original position and shows the options around it.

Before:

https://github.com/user-attachments/assets/59af5f4a-06ef-48ff-9a83-cbb81fe3663e

After:

https://github.com/user-attachments/assets/bc3e6b7b-7a02-483e-b631-b7a569e6df45

### DIFF
--- a/frontend/src/components/BlockPropertySections/TypographySection.ts
+++ b/frontend/src/components/BlockPropertySections/TypographySection.ts
@@ -49,12 +49,10 @@ const typographySectionProperties = [
 					const userFontOptions = filterOptions(
 						(userFonts.data || []).map((font: UserFont) => toOption(font.font_name as string)),
 						filterString,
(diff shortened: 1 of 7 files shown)


surajshetty3416 (changes requested):
@stravo1 I don't this selected option should be first... it should be at its place following the sequence but checked/highlighted

surajshetty3416 (approved):

PR #701, opened 2026-07-29
feat: dark-mode variant for background images

nish7x:
## What

Adds a **Dark Mode** option to the **Background** control's `+` menu, so a block can have a separate background (image / color / gradient) for dark mode — mirroring the dark variant already available on **Image** elements (`darkSrc`).

## How it works

- The dark background is stored as `dark:background*` styles (e.g. `dark:backgroundImage`), reusing the existing variant/state machinery — so the "+" menu, the variant row, and the popover editor all work with no new UI.
- On the published page it renders as `@media (prefers-color-scheme: dark)` **plus** `[data-theme="dark"]` / `[data-prefers-color-scheme="dark"]` selectors, matching how Builder already scopes dark mode (auto + manual toggle), instead of an invalid `:dark` pseudo-class.
- In the editor it previews live when the canvas dark-mode toggle is on (mirrors the image `darkSrc` swap).

## Changes

- **`BackgroundHandler.vue`** — expose the `dark` variant on the Background control; skip the hover-style transition for it (dark mode is a theme switch, not an interaction state).
- **`builder_page.py`** — `append_state_style` renders the `dark` state via a dark-mode media query + toggle selectors (`dark_mode_style` helper).
- **`BuilderBlock.vue`** — apply `dark:` styles in the canvas when `canvasDarkMode` is on.
- **`builderBlockCopyPaste.ts`** — absolutize the `dark:backgroundImage` URL on copy, like `darkSrc`.

## Testing

- Set a light + dark background on a block, published the page, and confirmed the background swaps between the light and dark values when the OS/browser color scheme flips.
- Verified the `+` menu lists **Dark Mode** alongside On Hover / On Active / On Focus, and that once set it appears as a variant row with its own background editor.

https://github.com/user-attachments/assets/947cc93b-50f1-483f-989a-56970488005a

### DIFF
--- a/builder/builder/doctype/builder_page/builder_page.py
+++ b/builder/builder/doctype/builder_page/builder_page.py
@@ -1322,8 +1322,27 @@ def append_state_style(style_obj, style_tag, style_class, device="desktop"):
 		if ":" in key:
 			state, property = key.split(":", 1)
 			css_property = camel_case_to_kebab_case(property)
-			style_string = f".{style_class}:{state} {{ {css_property}: {value}; }}"
+			declaration = f"{css_property}: {value};"
+			if state == "dark":
+				style_string = dark_mode_style(style_class, declaration)
+			else:
+				style_string = f".{style_class}:{state} {{ {declaration} }}"
 			style_tag.append(wrap_with_media_query(style_string, device))
+
+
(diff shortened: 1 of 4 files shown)


surajshetty3416:
@nish7x Please re-open once you address the review

PR #702, opened 2026-07-29
feat: render the canvas in a shadow root

stravo1:
Renders each canvas breakpoint inside a shadow root, so page CSS cannot reach the editor UI.

This is an alternative to the iframe approach on https://github.com/frappe/builder/pull/656

## Why a shadow root

A shadow root shares the viewport, the coordinate space, and the event loop with the editor. Rects, listeners, and drag events therefore need no translation across a frame boundary. Fonts need no extra handling, because `@font-face` is document scoped.

Two boundary rules need care, and `canvasShadowDom.ts` holds the helpers for both:

- `event.target` retargets to the host element. Use `getEventTarget`.
- `document.querySelector` does not descend into a shadow root. Use `queryCanvas`.

## Client scripts

A shadow root gives no JavaScript isolation, so page and block scripts run against a facade in `canvasScriptFacade.ts`. The facade points `document` and `window` at the breakpoint's root.

Run and Stop gate the scripts, from the command palette or `Ctrl+Shift+P`. A tap toggles the state. A hold previews and restores the earlier state on release. Stopping remounts the block tree, because a script can mutate the DOM in ways Vue cannot undo.

A shadow root resolves `@media` against the editor viewport, so every breakpoint would get the same answer. `pageScriptEmulation.ts` rewrites width-only queries to container queries, and the canvas host becomes a size container only when the CSS needs one.

## Styles

Each breakpoint adopts the editor stylesheets by reference, as constructable `CSSStyleSheet` objects. The document parses each sheet once, and every root shares it. A change to a sheet updates all adopters, so only a stylesheet that enters or leaves the document reassigns `adoptedStyleSheets`.

Block client script CSS goes to the breakpoint that registered it, not to every root.

## State styles

Hover, active and focus styles now render during preview. Before this branch the canvas dropped them. A block renders its regular styles inline, and an inline style cannot carry a pseudo class, so `BuilderBlock` deleted every `hover:` key.

`blockStateStyles.ts` emits them as real rules instead, which mirrors `append_state_style` in `builder_page.py`. An inline declaration outranks any rule, so each state declaration carries `!important`.

The rules apply during preview only. To select a block the user hovers it, and a block that restyles itself under the pointer is unusable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/components.d.ts
+++ b/frontend/components.d.ts
@@ -42,6 +42,7 @@ declare module 'vue' {
     BuilderBlock: typeof import('./src/components/BuilderBlock.vue')['default']
     BuilderBlockTemplates: typeof import('./src/components/BuilderBlockTemplates.vue')['default']
     BuilderCanvas: typeof import('./src/components/BuilderCanvas.vue')['default']
+    BuilderCanvasShadowRoot: typeof import('./src/components/BuilderCanvasShadowRoot.vue')['default']
     BuilderCommandPalette: typeof import('./src/components/BuilderCommandPalette.vue')['default']
     BuilderLeftPanel: typeof import('./src/components/BuilderLeftPanel.vue')['default']
     BuilderRightPanel: typeof import('./src/components/BuilderRightPanel.vue')['default']
--- a/frontend/src/builder.d.ts
+++ b/frontend/src/builder.d.ts
@@ -132,6 +132,9 @@ declare interface CanvasProps {
 	settingCanvas: boolean;
 	overlayElement: HTMLElement | null;
 	breakpoints: Breakpoint[];
+	scriptsRunning: boolean;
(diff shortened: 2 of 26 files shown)


surajshetty3416:

I'll verify a few things against the repo before writing the review.

`​`​`
grep -rn "highlightBlocksWithClientScripts" frontend/src
`​`​`

Let me check the actual usages.

**Tool calls (batched):**

- Grep `highlightBlocksWithClientScripts` in `/root/t3-review/builder/frontend/src`
- Grep `splitStylePrefix` in `/root/t3-review/builder/frontend/src/utils/helpers.ts` (with context)
- Grep `<BuilderCanvas` in `/root/t3-review/builder/frontend/src`
- Read `/root/t3-review/builder/frontend/src/utils/scriptSandbox.ts`

---

Issue #718, opened 2026-08-07
Upgrading to frappe-ui 1.0.0


netchampfaris:
frappe-ui is heading to a `1.0.0` tag. The breaking changes land as separate PRs. This issue is the single place that tracks all of them for Builder.

- Work top to bottom. Each section names the exact sites in this repo, with before/after.
- More sections get appended as more breaks land before the tag.
- Do not start until you bump the frappe-ui pin. Today it is `1.0.0-beta.21` in `frontend/package.json`.
- Full list of changes: [migration guide](https://github.com/frappe/frappe-ui/blob/main/docs/content/docs/migration.md).

| # | Change | PR | Fails at build? |
| --- | --- | --- | --- |
| 1 | `$socket` is no longer set by the plugin | frappe/frappe-ui#948 | No — throws at runtime |
| 2 | `Popover` v0 API removed; `Tooltip` `placement` and `#body` renamed (14 Popover, 9 Tooltip) | frappe/frappe-ui#956 | No — every break is silent |
| 3 | Dropdown `placement`, `{ group, items }` and `component:` rows removed (11 sites) | frappe/frappe-ui#957 | No — silent; `vue-tsc` flags them |

---

# 1. `$socket` is no longer set by the plugin

PR: frappe/frappe-ui#948

frappe-ui 1.0.0 removes `initSocket` and the `FrappeUI` plugin's `socketio` option. The plugin no longer creates a socket.io connection.

Builder is the only app that used the plugin's socket instead of building its own.

## 1.1 What breaks

`frontend/src/utils/realtimeHandler.ts:10`:

```ts
this.socket = getCurrentInstance()!.appContext.config.globalProperties.$socket;
```

`frontend/src/main.ts:17` calls `app.use(FrappeUI)` with no options. `socketio` used to default to `true`, so the plugin set `$socket` for you. It no longer does.

This is not a build failure. Reading `$socket` throws at runtime with a message naming the fix, so it will surface on the first page that opens a realtime connection rather than as `undefined` crashing somewhere else.

## 1.2 Fix

Create the connection in `frontend/src/main.ts` and assign it before `app.mount()`:

```ts
import { io } from 'socket.io-client'

const host = window.location.hostname
const port = window.location.port ? ':9000' : ''
const protocol = port ? 'http' : 'https'
const siteName = import.meta.env.DEV ? host : window.site_name

app.config.globalProperties.$socket = io(
  `${protocol}://${host}${port}/${siteName}`,
[...]

PR #733, opened 2026-08-13
refactor: upgrade to frappe-ui 1.0.0 (beta.52)

surajshetty3416:
Works through everything tracked in #718 as of today, pinned to `v1.0.0-beta.52` (the latest tag; all tracked PRs verified as ancestors).

### What's here

- **Socket** — the plugin no longer sets `$socket`; `src/socket.ts` builds the connection like other Frappe apps and `main.ts` assigns it before mount.
- **Popover / Tooltip v1** — all 14 + 9 listed sites, plus the sites that consume `ColorPicker`'s passthrough slot (`TextBlockBubbleMenu`, 4× `TokenManager`), which break identically. `#target`/`#body`/`#body-main` → `#trigger`/`#default` (+`bare` wherever the content brings its own panel — including the `#body-main` site in `AIPageGeneratorModal`, since the v1 `PopoverPanel` has visible chrome), `placement` → `side`/`align`, `arrowClass` → `:offset="12"`, `trigger="hover"` in `InputLabel` → v1 `Tooltip` with `#content`.
  - The trigger wires its own click now, so editable inputs living inside triggers (`ColorInput`, `ImageUploadInput`) are shielded with `@click.stop`, and `GradientEditor`'s draggable stops keep their "don't open after a drag" guard the same way.
- **Dropdown v1** — the 11 listed sites plus `PublishButton`, `PropertyLabel`, and the "New Script" dropdown in `PageClientScriptManager`. `placement` → `align`, `{ group, items }` → `{ group, options }`, and the slot's `open` is state now, so the old `@click="open"` handlers are gone. `PageActionsDropdown` forwards `align`; its triggers inside `router-link`s use `@click.stop.prevent` so opening the menu can't navigate.
- **Input refs** — the 2 listed `.el.focus()` sites plus two more the sweep found (`VersionHistory.vue`, `PagePersonaSurvey.vue`).
- **FileUploader** — `uploadArgs` flattened to props; `:private="false"` survives on both sites (verified against v1's `uploadOptions` plumbing). v1 no longer exposes `inputRef`, so `ImageUploadInput`'s programmatic open now reaches the hidden input through `$el`.
- **FeatherIcon** → lucide span in `TrackingDisabledNotice`.
- **Tailwind globs** — replaced with the `content` export. Note: `experimental/ListView` sits **outside** frappe-ui's published globs (only SpriteIcons/TextEditor/Calendar/Charts are covered), so an extra glob keeps ListView styled while we still use it.
- **Experimental imports** — `ListView` (3 files) and `AxisChart` (1 file) now import from `frappe-ui/experimental`.
- **`@framework/ui`** — dep added (`link:../../frappe/ui`), 12 files re-pointed. These use **deep subpath imports** (`@framework/ui/telemetry`, `@framework/ui/components/TrialBanner`) rather than the root barrel — see below.
- Follow-ups the v1 typings surfaced: Badge/Dialog dropped `orange`/`yellow` (→ `amber`), stale `frappe-ui > feather-icons` prebundle entry removed, `vue`/`frappe-ui`/`@vueuse/core` deduped in vite so `@framework/ui`'s peers resolve to this app's single copies.

### ⚠️ Upstream skew (why deep imports, and why this stays draft)

frappe/frappe's `ui` package hasn't caught up with two frappe-ui breaks yet:

1. `CodeEditorField.vue` still imports `frappe-ui/code-editor`, which was folded into the experimental barrel (frappe/frappe-ui#1043-era fold) — this makes the **root barrel** of `@framework/ui` unbuildable against beta.52, hence the deep imports.
2. `TrialBanner.vue` still imports the removed `FeatherIcon`. I verified the build with a two-line local patch to the frappe worktree (`FeatherIcon` → lucide span); that fix needs to land in frappe/frappe before this merges.

Once both are fixed upstream, the deep imports can collapse back to plain `@framework/ui` root imports.

### Verification

- `vite build` is green (in a bench-shaped worktree with frappe at develop + the TrialBanner patch above).
- `vue-tsc` baseline-diffed against develop/beta.21: **157 → 72 errors, zero new Builder-side errors** (the remainder are pre-existing plus frappe-ui's own experimental-barrel noise at beta.52).
- Not yet QA'd in the browser — that's why this is a draft. The behavioral hot spots to click through: color picker (focus-to-open, swatch toggle, token save), background/shadow handlers (open via row click now, not input focus), gradient stop drag vs click, page-card/list/route-tree "…" menus (must not navigate), AI modal style picker, image upload input + programmatic open from the store, search filter popover, token manager swatches, page-title popover (external-link icon must not toggle), publish dropdown.


surajshetty3416:

**Likely bugs**

- `RouteTreeNode.vue` — the slot signature changed to `<template v-slot>` but the `<Button icon="lucide-more-horizontal" …>` below it isn't shown as modified. Everywhere else with this pattern (`PageCard.vue`, `PageListItem.vue`, `PublishButton.vue`, `PageClientScriptManager.vue`) the accompanying `@click="open"` was removed in the same hunk. Confirm that button no longer references `open`, otherwise it's a runtime "open is not defined" on render.
- `ComponentUpdates.vue` — `<template #body="{ close }">` became `<template #default>` (no `close`), but the body still contains a `<Button …>` right below whose props are outside the hunk. If it still binds `close`, same problem.
- `ColorInput.vue` — `@click="toggle"` on the prefix swatch passes the `MouseEvent` as the first arg. `ColorPicker.vue` in this same PR explicitly guards against exactly that (`if (open instanceof Event) open = undefined`), but the `#trigger` slot now forwards frappe-ui's raw `toggle` via `v-bind="slotProps"`, so the guard is bypassed. With a truthy Event the swatch will only ever open, never close. Use `@click="() => toggle()"`.
- `ImageUploadInput.vue` — `fileUploaderRef.value?.$el.querySelector(...)`: if `FileUploader` has more than one root node, `$el` is the fragment anchor (a comment/text node) and `.querySelector` is not a function → TypeError. At minimum `?.$el?.querySelector?.(...)`; better, have `FileUploader` expose the input or use a template-scoped `document` lookup.
- `frontend/src/socket.ts` — `socket.io-client` is imported but not declared in `frontend/package.json`; it currently resolves only transitively. Also `":9000"` is hardcoded rather than read from the bench's socketio port, so it breaks on any site with a non-default port. And it's now initialized unconditionally in `main.ts` even on preview/public routes.

**Worth verifying**

- `TextBlockBubbleMenu.vue` — `@open="suppressColorOptions"` on `ColorPicker`, but the diff only shows `ColorPicker` consuming `@update:open` internally; nothing shows it re-emitting `open`. If it doesn't, the listener falls through as a fallthrough attr and never fires, silently regressing the "don't auto-open the option list" workaround. Also check whether `selectedColor`/`setTextColor` are now unused after deleting the stray `<template>` block.
- `MoreStylesPanel.vue` (`typeof v === 'string' ? v : null`) and `PageClientScriptManager.vue` (`typeof value === 'string' && attachScript(value)`) — if the v1 `Autocomplete`/`Combobox` emit the option object rather than a bare value, these casts turn a type error into a silent no-op (property never added, script never attached). Confirm the emitted shape rather than narrowing to `string`.
- `tsconfig.json` maps `frappe-ui` → `../frappe-ui/src/index.ts` (submodule) while `package.json` bundles the published `1.0.0-beta.52`. Unless the submodule commit is exactly beta.52 and it's yarn-linked, type-checking and the actual bundle diverge.
- `tailwind.config.js` adds `frappeUIContent` but nothing covers `@framework/ui` (`../../frappe/ui/src`). Any Tailwind classes used by `TrialBanner` and friends won't be generated unless that app ships its own compiled CSS.

The rest (class reordering, `orange`→`amber` themes, `items`→`options`, `placement`→`side`/`align`, `frappe-ui/frappe`→`@framework/ui/telemetry`) looks like a consistent, correct mechanical migration; the `@click.stop`/wrapper-div workarounds around the new merged trigger handlers are reasonable and the added comments explain why.

---

PR #745, opened 2026-08-13
fix(page-data-script): make translation actually callable as `_(...)`

### LINKED ISSUE #626: `frappe._()` does not work inside a Builder Page Data Script, even though the function is included in the sandbox globals.
## Description
`frappe._()` does not work inside a Builder Page Data Script, even though the function is included in the sandbox globals.

## Steps to reproduce
1. Create a Builder Page
2. Add a **Page Data Script** containing: `data.update({"text": frappe._("Supplier")})`
3. Bind a text block's Data Key to `text`
4. Preview the page

## Expected behavior
The translated string appears (e.g. `"Lieferant"` for German).

## Actual behavior
`'str object' has no attribute 'get'` — or an empty string is rendered.

## Root cause
**File:** `builder/builder/utils.py` — function `get_safer_globals()`

The `_` function is only available inside the `frappe` namespace (line 241):

```python
frappe=NamespaceDict(
    ...
    _=frappe._,   # line 241 — only here
    ...
)
```

When the sandbox script calls `frappe._("text")`, RestrictedPython compiles the attribute access `frappe._` into `_getattr_(frappe, "_")`. The `_getattr_` guard (from `frappe.utils.safe_exec._validate_attribute_read`) blocks **all** attribute names starting with `_`:

```python
# frappe/utils/safe_exec.py line 571:
if name.startswith("_"):
    raise AttributeError(f'"{name}" is an invalid attribute name because it starts with "_"')
```

This means `frappe._()` can never succeed — the guard kills it before the function is ever reached.

## Fix
Add `_=frappe._` to the **top-level** of the `out` NamespaceDict, matching how Frappe's own `get_safe_globals()` works:

## Environment
- Frappe/ERPNext version: v16
- Builder version: 1.24.7

suparikoli:
Fixes #626.

The issue's diagnosis is right and I ran it down the same way to be sure before touching anything: `frappe._("Supplier")` in a Page Data Script never gets anywhere near `frappe._`. `frappe._` is an attribute read, and RestrictedPython compiles that to `_getattr_(frappe, "_")`. That guard function is `_getattr_for_safe_exec` in `frappe/utils/safe_exec.py`, and it calls `_validate_attribute_read`, which rejects any attribute name starting with `"_"` unconditionally, before it even looks at what the attribute would resolve to:

```python
def _validate_attribute_read(object, name):
    ...
    if name.startswith("_"):
        raise AttributeError(f'"{name}" is an invalid attribute name because it starts with "_"')
```

So nesting `_=frappe._` differently inside the `frappe` namespace wouldn't have changed anything - `frappe._(...)` can't work regardless of where `_` is defined, because it's the attribute access itself that's rejected, not a lookup failure.

A bare name reference (`_("Supplier")`, no dot) compiles to a plain global lookup, which doesn't go through `_getattr_` at all - RestrictedPython only wraps attribute access, not name resolution. That's exactly what frappe's own `render_safe_globals()` (the base sandbox builder's `get_safer_globals()` extends) already relies on: it puts `_=frappe._` at the top level of `out`, not under `out.frappe`. Builder's version has it nested one level down, which is the actual gap. Moved it to match.

Added a test in `test_utils.py` alongside the existing `execute_script` ones: runs `data.text = _("Supplier")` through `execute_script()` and checks it resolves to `"Supplier"` (no translation is loaded for that string in a test run, so it comes back unchanged, which is expected), and separately asserts that `frappe._("Supplier")` still raises - want that locked in so nobody "fixes" this again later by moving `_` back under `frappe`, which would look reasonable but doesn't actually work.

Ran `ruff check` and `ruff format --check` against both changed files - clean. Didn't have a live site to run the actual test against; the reasoning above is what I'm relying on instead, traced against current `frappe/utils/safe_exec.py` and `render_safe_globals()` rather than assumed.

### DIFF
--- a/builder/builder/tests/test_utils.py
+++ b/builder/builder/tests/test_utils.py
@@ -139,6 +139,18 @@ class TestBuilderUtils(FrappeTestCase):
 		execute_script("data.sum = a + b", {"data": data, "a": 2, "b": 2}, "test.py")
 		self.assertEqual(data.sum, 4)
 
+	def test_execute_script_can_translate(self):
+		# frappe._(...) can't work in a Page Data Script: it's an attribute
+		# read, and the sandbox's _getattr_ guard rejects any name starting
+		# with "_" before ever getting to what "_" resolves to. "_" has to be
+		# called bare - see get_safer_globals(), and #626.
+		data = frappe._dict({})
+		execute_script('data.text = _("Supplier")', {"data": data}, "test.py")
+		self.assertEqual(data.text, "Supplier")
+
+		with self.assertRaises(Exception):
+			execute_script('data.text = frappe._("Supplier")', {"data": data}, "test.py")
+
 	@patch("builder.utils.is_safe_exec_enabled", return_value=False)
 	@patch("frappe.utils.safe_exec.is_safe_exec_enabled", return_value=False)
 	def test_execute_script_with_enabled_server_script(self, *args):
--- a/builder/builder/tests/test_utils.py
+++ b/builder/builder/tests/test_utils.py
@@ -140,14 +140,11 @@ class TestBuilderUtils(FrappeTestCase):
 		self.assertEqual(data.sum, 4)
 
 	def test_execute_script_can_translate(self):
-		# frappe._(...) can't work in a Page Data Script: it's an attribute
-		# read, and the sandbox's _getattr_ guard rejects any name starting
-		# with "_" before ever getting to what "_" resolves to. "_" has to be
-		# called bare - see get_safer_globals(), and #626.
 		data = frappe._dict({})
 		execute_script('data.text = _("Supplier")', {"data": data}, "test.py")
 		self.assertEqual(data.text, "Supplier")
 
+		# the sandbox rejects `frappe._`, only the bare name works (#626)
 		with self.assertRaises(Exception):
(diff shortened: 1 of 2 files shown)


surajshetty3416:

Change looks correct. `_` is the one underscore-prefixed name RestrictedPython's `check_name` explicitly permits, and frappe's `_getattr_` guard does reject the `frappe._` attribute path at runtime, so both the fix and the comment are accurate. Exposing `frappe._` adds no new capability beyond translation lookup.

Two minor test nits (non-blocking):

- `builder/builder/tests/test_utils.py:143` — `_("Supplier")` only equals `"Supplier"` when the test site's language is `en`. If a site under test has a non-English default lang with a Supplier translation, this breaks. Wrapping in `frappe.set_user_lang`/`with frappe.change_site_config`-style lang pinning, or asserting on a string with no translation entry, would make it deterministic.
- `builder/builder/tests/test_utils.py:147` — `assertRaises(Exception)` is broad enough to pass for the wrong reason (e.g. if `frappe` ever stopped being in globals, you'd get a `NameError` and the test would still be green). `assertRaises(SyntaxError)` matches what the guard actually raises and pins the intended behavior.

---

surajshetty3416:
@ankush is this safe?

PR #747, opened 2026-08-14
feat: recently used colors in the color picker palette

SmrutiSawant:
### What

The color picker's swatch row was a hardcoded list of 8 colors. It now surfaces the recently used colors, and the color popovers
are sized consistently.

### Changes

**Recently used colors**

- Last 8 colors persist to `localStorage`, shared across every picker instance in the tab.
- The static palette backfills unused slots, so the row is never empty.
- Colors are banked when the popover closes, not on every drag frame.
- Deduped, so a color can't occupy two slots.
Before :

https://github.com/user-attachments/assets/3b1c7fe4-d1c2-4d69-9c25-482d96e2a370

Now:

https://github.com/user-attachments/assets/ffde152e-5da0-4f63-99f9-02921b27e4a0

**Consistent popover width**

- Background and picker popovers each sized themselves independently. All are now `w-52` (208px), which fits the swatch row without leaving dead space.
Before :
before

Now : 
now

### DIFF
--- a/frontend/src/components/BackgroundHandler.vue
+++ b/frontend/src/components/BackgroundHandler.vue
@@ -1,5 +1,5 @@
 <template>
-	<Popover placement="left" class="!block w-full" :offset="25">
+	<Popover placement="left" class="!block w-full" :offset="25" @update:open="handlePopoverToggle">
 		<template #target="{ togglePopover }">
 			<div class="flex w-full items-center justify-between" @focusin="updateActiveState">
 				<StylePropertyControl
@@ -47,6 +47,7 @@
 				<!-- Color Tab -->
 				<div v-if="activeTab === 'color'" class="w-full space-y-4">
 					<ColorPicker
+						ref="colorPickerRef"
 						renderMode="inline"
 						:modelValue="backgroundColor"
 						:showInput="true"
@@ -178,6 +179,12 @@ const BackgroundInput = defineComponent({
 });
 
 const activeState = ref<string | null>(null);
+const colorPickerRef = ref<InstanceType<typeof ColorPicker> | null>(null);
+
+// the picker renders inline here, so this popover owns the "closed" moment
+const handlePopoverToggle = (open: boolean) => {
+	if (!open) colorPickerRef.value?.commitRecentColor();
+};
 
 const updateActiveState = (e: FocusEvent) => {
 	const target = e.target as HTMLElement;
--- a/frontend/src/components/BackgroundHandler.vue
+++ b/frontend/src/components/BackgroundHandler.vue
@@ -34,7 +34,7 @@
 		</template>
 		<template #body>
 			<div
-				class="background-popover-body w-64 rounded-lg border border-outline-gray-2 bg-surface-base p-3 shadow-xl">
+				class="background-popover-body w-52 rounded-lg border border-outline-gray-2 bg-surface-base p-3 shadow-xl">
 				<TabButtons
 					:options="[
 						{ label: '', value: 'color', icon: 'lucide-droplet' },
--- a/frontend/src/components/Controls/ColorPicker.vue
+++ b/frontend/src/components/Controls/ColorPicker.vue
@@ -5,7 +5,7 @@
 		:placement="placement"
 		:offset="offset"
 		:portal-to="portalTo"
-		@update:open="isOpen = $event"
+		@update:open="handleOpenChange"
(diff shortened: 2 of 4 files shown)


surajshetty3416:

I'll verify a couple of things against the actual file before writing the review.

**Bash**
```
cd /root/t3-review/builder && grep -n "currentColor" frontend/src/components/Controls/ColorPickerContent.vue
```

**Result**
```
92:let currentColor: CSSColorValue | null = null;
...
238:  const updateColor = () => {
239:    const alphaValue = a < 1 ...
```

**Read** `frontend/src/components/Controls/ColorPickerContent.vue` (lines 80–140, 230–300)

**Read** `frontend/src/components/Controls/ColorPicker.vue`

---

Review:

**`ColorPickerContent.vue` — `lastColor` watch is `immediate`, so merely opening a picker banks a color**

```ts
watch(modelColor, (color) => color && (lastColor = color), { immediate: true });
```

Combined with `onBeforeUnmount(commitRecentColor)`, opening any color popover and closing it without touching anything writes the block's *existing* color to the recents list, pushing genuinely-used colors out of the 8-slot window. This also fires for the second/third pickers a user merely glances at. GradientEditor deliberately avoids this (`lastGradient` starts `null` and is only set in `applyPreset`/`emitUpdate`). Suggest matching that: drop `immediate: true` and only assign `lastColor` from `selectColor`/`handleInputChange`/`updateColor`, i.e. from actual user edits rather than from prop sync.

**`GradientEditor.vue` — recents can evict the presets entirely**

`gradientSwatches` caps at `presets.length` and puts recents first, and `recentGradients` is also sliced to `presets.length`. Once the user applies N distinct gradients, every built-in preset disappears permanently with no way back. The color picker has the same shape but it's less severe there since custom colors are cheap to retype; a gradient preset isn't. Consider reserving a few slots (e.g. cap recents at 3–4 and let presets fill the rest), which is also how the row reads in the comment ("presets seed whatever space is left over").

**`GradientEditor.vue` — `stopPickerRefs` is keyed by index and never pruned**

```ts
const stopPickerRefs: Record<number, any> = {};
```

When a stop is removed, the entries for the shifted indices are reassigned but the entry for the old highest index lingers (Vue sets it to `null` on unmount, so it's harmless today). The real point is that this whole mechanism is redundant: `ColorPickerContent` already commits in its own `onBeforeUnmount`, and the stop's `ColorPicker` lives inside the popover's `#body`, so closing the popover unmounts it anyway. The same applies to `handlePopoverToggle` in `BackgroundHandler.vue` and `handleOpenChange` in `ColorPicker.vue` — by the time `update:open(false)` fires, `contentRef`/`colorPickerRef` may already be `null`, and the unmount hook is what actually does the work. Dropping the three explicit call sites would remove the `any`-typed index map and the ref plumbing without changing behavior. If you keep them, `commitRecentColor` is at least idempotent (re-committing the same color is a no-op reorder), so there's no double-entry bug.

**`selectColor` bypassing `updateColor()` — verify alpha isn't dropped**

```ts
currentColor = color;
emit("update:modelValue", color);
```

`currentColor` is a plain `let`, so the assignment is correct. But `updateColor()` previously derived the emitted value from the hue/alpha selector positions, and it's the only thing that reconciles them. Since swatches are 6-digit hex, selecting a swatch while the alpha slider sits below 1 now emits a fully opaque color while the alpha selector still renders at its old position — the UI and the emitted value disagree until the next drag. Worth either resetting the alpha selector in `setSelectorPosition` for opaque input or appending the current alpha to the emitted swatch value.

**Minor**

- `normalizeColor` is applied when comparing and when storing, but `defaultColors` relies on being hand-lowercased to match. A stray uppercase literal added later would silently produce a duplicate swatch. Cheap fix: `defaultColors.map(normalizeColor)` at definition, or normalize inside the `colors` computed's `palette.push`.
- `useStorage` with untrusted `localStorage` content: `recentColors`/`recentGradients` are read back and interpolated into `:style="{ background: swatch.gradient }"`. Vue's style binding won't execute script from a string, so this isn't XSS, but a corrupted array entry (non-string) would throw in `normalizeColor(...).trim()`. A `typeof c === "string"` filter when reading would make it robust.

Nothing else blocking — the `flush: "sync"` reasoning is correct and well-commented, and the dedupe logic in both computeds is sound.

---

surajshetty3416 (approved):

PR #748, opened 2026-08-14
fix(a11y): add alt text to images missing it

swe-sanad:
## Problem

Four images had no `alt` attribute.

## Fix

Preview thumbnails of a named item get a real alt = that item's title (block-template preview, page card/list previews) — matching frappe/insights' own convention for the same kind of preview card. `ViewerAvatars.vue` has no visible name text next to the avatar (only a hover tooltip, which isn't a reliable accessible-name source), so it gets a real alt too.

### DIFF
--- a/frontend/src/components/BuilderBlockTemplates.vue
+++ b/frontend/src/components/BuilderBlockTemplates.vue
@@ -22,7 +22,7 @@
 							:class="{
 								'w-14': !blockTemplate?.preview_width || blockTemplate?.preview_width == 1,
 							}">
-							<img :src="blockTemplate.preview" class="pointer-events-none" />
+							<img :src="blockTemplate.preview" :alt="blockTemplate.template_name" class="pointer-events-none" />
 						</div>
 						<p class="text-wrap text-center text-sm text-ink-gray-6">
 							{{ blockTemplate.template_name }}
--- a/frontend/src/components/PageCard.vue
+++ b/frontend/src/components/PageCard.vue
@@ -9,6 +9,7 @@
 				width="250"
 				height="140"
 				:src="page.meta_image || page.preview"
+				:alt="page.page_title || page.page_name"
 				onerror="this.src='/assets/builder/images/fallback.png'"
 				class="aspect-video w-full overflow-hidden rounded-md object-cover shadow dark:border dark:border-outline-gray-1" />
 			<div class="flex items-center justify-between border-outline-gray-2">
--- a/frontend/src/components/PageListItem.vue
+++ b/frontend/src/components/PageListItem.vue
@@ -10,6 +10,7 @@
 					width="140"
 					height="82"
 					:src="page.meta_image || page.preview"
+					:alt="page.page_title || page.page_name"
 					onerror="this.src='/assets/builder/images/fallback.png'"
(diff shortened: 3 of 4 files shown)


surajshetty3416:

The change is correct and low-risk — four `alt` attributes added, all bound to values that exist alongside the images already rendered in the same templates. No bugs or security concerns.

One optional a11y nit: in `BuilderBlockTemplates.vue`, `PageCard.vue`, and `PageListItem.vue` the same text (`template_name` / `page_title`) is already rendered as visible text immediately adjacent to the image, so screen readers will announce it twice. For purely decorative previews like these, `alt=""` is the more standard treatment. Same applies to `ViewerAvatars.vue`, where `alt` now duplicates the `title` on the identical element. Not blocking — duplicated alt text is still better than missing alt text.

---

surajshetty3416 (commented):
Thanks for this! The avatar change is good as is: no name is shown next to it, so the alt is the only accessible name.

For the page and block template previews, the title is already visible right next to the image (and inside the same link for pages), so an alt with the title makes screen readers announce it twice. These previews are decorative, so `alt=""` fits better. Suggestions inline.

The branch also conflicts with `develop` now. After applying the suggestions, please merge `develop`. For `PageCard.vue`, keep develop's version (the image is now wrapped in a `relative` div with `PageStatusBadge`) and add `alt=""` to that `img`.

`Controls/ImageFocusInput.vue` landed after this PR was opened and has an `img` without alt too. It's the focal point picker preview, so `alt=""` works there as well.

frontend/src/components/PageCard.vue
```diff
@@ -9,6 +9,7 @@
 				width="250"
 				height="140"
 				:src="page.meta_image || page.preview"
+				:alt="page.page_title || page.page_name"
```
surajshetty3416: The page title is shown right below inside the same link, so this reads as "Title Title".

```suggestion
				alt=""
```

frontend/src/components/PageListItem.vue
```diff
@@ -10,6 +10,7 @@
 					width="140"
 					height="82"
 					:src="page.meta_image || page.preview"
+					:alt="page.page_title || page.page_name"
```
surajshetty3416: Same as the card: the title is visible inside the same link.

```suggestion
					alt=""
```

frontend/src/components/BuilderBlockTemplates.vue
```diff
@@ -22,7 +22,7 @@
 							:class="{
 								'w-14': !blockTemplate?.preview_width || blockTemplate?.preview_width == 1,
 							}">
-							<img :src="blockTemplate.preview" class="pointer-events-none" />
+							<img :src="blockTemplate.preview" :alt="blockTemplate.template_name" class="pointer-events-none" />
```
surajshetty3416: `template_name` is rendered right under the preview.

```suggestion
							<img :src="blockTemplate.preview" alt="" class="pointer-events-none" />
```

frontend/src/components/PageCard.vue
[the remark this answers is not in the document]
surajshetty3416: Agreed, the title is visible next to each preview. Left `alt=""` suggestions on all three; the avatar keeps its alt since it has no visible name.

PR #753, opened 2026-08-15
chore: sync translations from crowdin

surajshetty3416:

PR #756, opened 2026-08-16
chore: sync translations from crowdin

surajshetty3416:


mardausdennis:
German translations created on or after 2026-08-08 are not being exported.

`builder/locale/de.po` on `develop` carries 232 translations. That is exactly the set created in Crowdin before 2026-08-08. Crowdin holds 881 for the same file. The 117 created on 08-08 and 08-13 never landed, and neither did the 532 I added on 08-14.

The last German export was a454ecc on 2026-08-13, which took the file from 51 to 232, so the pipeline itself works.

Other languages are unaffected. #753 carried Bosnian and Hungarian, this one carries Hungarian.

Approval state does not explain it either. German has 7 approved strings, Hungarian has 2.

Is there something I can do from the translator side, or does this sit in the integration?

surajshetty3416:
@mergify update

PR #759, opened 2026-08-17
feat(ai): give Bob one voice across models

surajshetty3416:
## Why

The agent prompt had functional rules but no persona, so each provider's default tone leaked through: one model gushes, another apologises in loops, a third answers like a press release. Users who switch models mid-project (the picker sits right in the composer) heard the seam.

## What

**One persona, defined once.** Bob's voice now lives in a shared `Prompts.VOICE` constant (alongside `STYLING_RULES`/`MEMORY_RULES`) so future agents and orchestrators reuse it verbatim. The persona: a warm, sharp web designer working beside the user at the canvas.

- Warmth is attention, not energy: acknowledge what the user is trying to do in specifics, then act. Sorry is said once, and only when Bob actually caused the problem.
- Curious about the user's world: carry the specifics they hand over (the studio's name, the signature dish) into the work; ask only when one thing would genuinely change the build, never an interrogation.
- Oriented at the end goal: read each ask as a step toward bookings, orders or signups, say so when it shapes the work, and offer one concrete next step when it is earned. Generic "anything else?" closes are banned.
- Opinionated with reasons: a direct "should I?" gets an answer in words, never a dodge into an options card or a silent change.
- No sycophancy, no self-praise ("stunning", "perfect"), no emoji, at most one exclamation at a true milestone, honest in passing about the running model when asked (the picker shows it anyway), and replies in the user's language.

**No weak-model handholding.** The persona is rules-only (no few-shot example pair), and the loop's response-salvage machinery is removed: text-emitted tool-call salvage, the card-as-text and options-as-prose corrective rounds, the no-op-claim guard and its finish-time backstop, the incomplete-build correction, and YAML-as-content recovery. A model's reply now stands as sent; quality differences between tiers are allowed to show. Transport robustness (json_repair of streamed args, provider compat, stream retries, empty-reply fallbacks) stays.

**The canvas follows the build.** During a generation the interesting part is the page's growing bottom edge, and it scrolls out of the viewport within seconds; users were panning manually to watch their own build. A new `useBuildFollow` composable auto-pans the camera (pan only, never zoom): streaming keeps the leading edge parked just above the viewport bottom with an eased glide, and each agent edit batch pans to the last touched block. Any user camera move (wheel, grab-drag, zoom, fit) pauses following so Bob never fights the user for the camera; panning back to the leading edge, or a new build, resumes it. The resume check credits the edge's position before the latest chunk's growth, since the page grows again between the user's pan and the next render.


surajshetty3416:
@mergify update

PR #766, opened 2026-08-17
chore: update POT file

mardausdennis:
Regenerated `builder/locale/main.pot` from `develop`. The committed file was last updated on 2026-08-13, so it predates the strings wrapped in #746 and #750 and the AI rewrite in #754.

This is the first successful extraction since #758. I ran the same steps as `.github/helper/update_pot_file.sh` on a plain `ubuntu-latest` runner, which also confirms that the fix in #758 works: `bench generate-pot-file --app builder` now completes instead of raising `RuntimeError: object is not bound`. The run took 1m42s.

The result is 1165 strings, up from 929.

- 271 added, mostly from #746 and #750 plus the settings, domains, redirects and token manager work
- 35 removed, all from the AI panel rewrite, where the old strings no longer exist in the source

One aside while comparing: `Drop image to attach` disappears from the POT because it is now plain template text in `frontend/src/components/BuilderAIChatPanel.vue` rather than a `__()` call. It looks like the wrapper was lost in the rewrite. Not touched here, since this PR is only the regenerated file, but it is worth a look.

The next scheduled run is Sunday. This PR just gets the current strings into the catalog now, so translators are not a week behind.

### DIFF
--- a/builder/locale/main.pot
+++ b/builder/locale/main.pot
@@ -7,14 +7,63 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Frappe Builder VERSION\n"
 "Report-Msgid-Bugs-To: suraj@frappe.io\n"
-"POT-Creation-Date: 2026-08-13 08:13+0553\n"
-"PO-Revision-Date: 2026-08-13 08:13+0553\n"
+"POT-Creation-Date: 2026-08-18 15:53+0000\n"
+"PO-Revision-Date: 2026-08-18 15:53+0000\n"
 "Last-Translator: suraj@frappe.io\n"
 "Language-Team: suraj@frappe.io\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Generated-By: Babel 2.16.0\n"
+
+#: frontend/src/components/Modals/TokenManager.vue:965
+msgid "({0} invalid entries skipped)"
+msgstr ""
+
+#: frontend/src/components/Modals/TokenManager.vue:965
+msgid "({0} invalid entry skipped)"
+msgstr ""
+
+#: frontend/src/components/RouteTreeNode.vue:91
+#: frontend/src/components/RouteTreeView.vue:41
+msgid "({0} remaining)"
+msgstr ""
+
+#: frontend/src/components/Modals/TokenManager.vue:966
+msgid "({0} standard token skipped)"
+msgstr ""
+
+#: frontend/src/components/Modals/TokenManager.vue:966
+msgid "({0} standard tokens skipped)"
+msgstr ""
+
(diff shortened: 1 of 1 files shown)


mardausdennis:
The failing Server check is not caused by this PR. It looks like `develop` is currently broken against `frappe/develop`.

Evidence:

- This PR changes exactly one file, `builder/locale/main.pot`. A gettext template cannot affect block styles or script escaping.
- `develop` is still at `6a26ad71`, and the push run for that same commit passed all 275 tests at 10:24 UTC today.
- `frappe/frappe` merged frappe#41626 (`fix/sanitize-json-content`) at 10:59 UTC, which touches `frappe/utils/html_utils.py`.
- Builder CI installs frappe from `develop` unpinned, so every run after 10:59 gets the new sanitizer.
- I re-ran this PR with an identical tree at 14:07 and again at 16:40 UTC. Both fail the same two assertions, so it is not flaky.

The two failures both involve content that passes through HTML sanitization:

```
AssertionError: 'var(--new-brand, #eee)' not found in
'[{"blockId": "root", "baseStyles": {"color": "var(--new-brand)", ...
```

```
AssertionError: '<\/script><p>Component Script</p>' not found in '<!DOCTYPE html> ...
```

The design token fallback is gone and the `</script>` escaping is no longer applied.

As far as I can tell this PR is just the first one tested after that frappe merge, since `develop` has not been pushed since 10:24. Worth checking independently of this PR.

mardausdennis:
Root cause, and it is not a frappe regression.

frappe#41626 deliberately removed the JSON bypass in `sanitize_html`:

```diff
 	if not always_sanitize:
-		if is_json(html):
-			return html
-
 		if not bool(BeautifulSoup(html, "html.parser").find()):
 			return html
```

The old docstring already called that bypass risky, and the new tests assert that `{"x": "<script>alert(1)</script>"}` no longer passes through. The change is intentional and correct.

What it means here: `builder_page.blocks` and `draft_blocks` are `Long Text`, and the script and html fields are `Code`. Their content is block JSON carrying markup, so it used to take the JSON bypass and reach the database untouched. Now nh3 rewrites it. That is why the design token fallback `, #eee` disappears, and why the `</script>` escaping no longer appears: Builder escapes on render, so the sanitizer was never the security boundary for these fields. It was simply never reached before.

Two directions, and the choice is a security call I would rather not make on your behalf:

1. Set `ignore_xss_filter: 1` on the affected fields. Builder already does this for `route`, so the mechanism is established. It makes the previous behaviour explicit instead of accidental.
2. Move `blocks` and `draft_blocks` to the `JSON` fieldtype, matching `builder_component.block`. Cleaner semantically, but a schema change.

I can send either as a PR if you say which direction you want. I can also verify the candidate first by running `server-tests` on a fork before proposing it, so it would not arrive untested.

Independent of this PR, every Builder PR tested after 10:59 UTC today will hit the same two failures.

surajshetty3416 (approved):

PR #773, opened 2026-08-20
fix: AI design flow session review fixes

surajshetty3416:
Fixes from reviewing real Bob sessions, plus two small design-flow trims.

**Attached images survive past the turn they were sent on.** A pasted reference image rode the request as a data URI, reached the model once, and vanished: session context replay was text-only, and the generation step only saw the current turn's attachment plus brief markers. A user who pasted a design and then answered the flow's questions got a page built blind, with the reference ignored. Now:
- the attachment is saved as a site file (`Home/Builder Uploads`) and the message metadata stores its URL instead of the one-turn data URI
- the newest attached images ride their user turns again for vision models, in both the agent loop and the generation step (legacy data-URI metadata still replays as-is)
- the live turn names the saved URL so the agent can carry it into a brief as a `REFERENCE IMAGE:` line
- the design flow treats an attached page design as the approved direction: no layout/colour questions it already answers

**Enter on a mid-card input advances the step.** On a multi-question card, Enter submitted the whole card from any step, skipping the remaining questions. It now answers the current step; only the last step's Enter submits.

**Layout direction options drop the palette pill.** The colours step already settles the palette with role slots, so every direction option wore the same swatch strip, saying nothing about layout.

**Remove the portfolio skill.** The playbook wasn't lifting portfolio builds above the base generation rules; drop the registry, the `generate_page` skill parameter, the injection, and the timeline label.

Tests: `builder.ai.test_attached_images` covers the file save round-trip, replay selection and inlining; existing artifact/cache-marker suites pass.


surajshetty3416:
Both review findings addressed in c9c096ee: attachments now save as private files (the replay and brief inline paths read /private/files/ through permissioned lookups, so nothing needed a public URL), and the save moved below the session busy check so a 429'd request persists nothing.

surajshetty3416:
Addressed the new P1: `read_site_image` now checks File read permission before inlining a private file, so a brief-written `/private/files/` path only resolves when the requesting user (the background job runs as the enqueuing user) may read it. Covered by a denial test plus an owner-path test.

On the busy-check race: the fast check was always advisory; in the overlap window the user message is still appended, so the saved attachment is referenced by that message's metadata either way — nothing ends up orphaned. The queued worker's atomic run lock stays the authoritative guard, as before this PR.

PR #787, opened 2026-08-28
fix: block traversal segments in standard page export paths

surajshetty3416:
The semgrep `frappe-security-file-traversal` rule fires on the standard page exporters (`export_page_as_standard`, `export_client_scripts`), which write fixtures into an installed app's `builder_files/` directory.

Both are gated behind `frappe.conf.developer_mode` and the path segments come from `export_dir_name()`, which scrubs the record name and replaces slashes. It did not, however, reject `.` or `..` as a whole segment, so a record named `..` would write one directory above its export root.

`template_sync.py` already had `safe_segment()` for exactly this. Moved it into `builder/utils.py` (which `template_sync` already imports from), routed `export_dir_name()` through it, and marked the audited writes `# nosemgrep`, matching how the template exporter annotates its own.

Tested with `test_standard_page_sync` (28) and `test_utils` (21), all passing.

### DIFF
--- a/builder/builder/tests/test_utils.py
+++ b/builder/builder/tests/test_utils.py
@@ -14,6 +14,7 @@ from builder.utils import (
 	copy_img_to_asset_folder,
 	escape_single_quotes,
 	execute_script,
+	export_dir_name,
 	extract_components_from_blocks,
 	get_builder_page_preview_file_paths,
 	get_template_assets_folder_path,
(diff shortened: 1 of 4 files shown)


surajshetty3416:
Added regression coverage for the guard in `test_utils` (7d8211d): normal names, separator neutralisation (`../../etc/passwd` -> `.._.._etc_passwd`), and `ValidationError` on the empty/`.`/`..` segments.

PR #796, opened 2026-09-03
fix: two keyboard shortcut bugs

SmrutiSawant:
**What**

Ctrl+Shift+C leaked a listener per press copy-block-styles called useStorage() inside action(). VueUse attaches cleanup to Vue's active effect scope; a keypress handler has none, so the cleanup was silently dropped. Now hoisted to module scope

Ctrl+Shift+L was bound twice
The "Highlight Blocks with Client Scripts" hold binding collided with the Layers tab (highlightBlocksWithClientScripts was written, never read). Removed the binding and the dead store field.

Screenshot 2026-09-03 120028

### DIFF
--- a/frontend/src/components/Commands/index.ts
+++ b/frontend/src/components/Commands/index.ts
@@ -85,6 +85,11 @@ const builderStore = useBuilderStore();
 const pageStore = usePageStore();
 const canvasStore = useCanvasStore();
 
+// module scope, not inside the command action: useStorage subscribes to storage
+// events and registers its cleanup on the active effect scope. A key handler has
+// none, so calling it per keypress would leak a subscription every time.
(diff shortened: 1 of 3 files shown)


frontend/src/components/Commands/index.ts
```diff
@@ -85,6 +85,11 @@ const builderStore = useBuilderStore();
 const pageStore = usePageStore();
 const canvasStore = useCanvasStore();
 
+// module scope, not inside the command action: useStorage subscribes to storage
```
stravo1, line 88: we can make the comment more concise.

stravo1 (approved):

surajshetty3416 (approved):

PR #797, opened 2026-09-04
fix: match font dropdown width to its input field

SmrutiSawant:
**Problem**
The font dropdown was pinned to a 240px minimum width, unlike other property dropdowns.
image

**Solution**
Remove the fixed minimum width and use a hover marquee to reveal truncated labels.
When enabled, long option labels scroll horizontally into view on hover, preserving access to the full name without increasing the dropdown width.
Recording 2026-09-04 154814


surajshetty3416:
Not needed as of now. 

Current one is better for quick glance preview.

PR #839, opened 2026-09-17
fix: normalize nested block selections

stravo1:
## Problem

A multi-selection can contain a block and a block nested in it. Copy and paste then duplicate the nested block.

Before:

https://github.com/user-attachments/assets/f99c5919-9612-44b0-9516-4c9c0d2f0976

After:

https://github.com/user-attachments/assets/6e51f185-4e02-49a9-98c7-d4f9b3938a74

## Changes

- A selection keeps only the outermost blocks. `removeNestedBlocks` in `useBlockSelection` removes each block that has a selected ancestor. It uses parent links, not tree lookups.
- When the user adds a block to a multi-selection, remove its selected descendants. When a selected block contains the block that the user adds, keep the selection as it is.
- A marquee selection uses the same rule. An additive marquee merges the old selection with the blocks under the marquee, then removes the nested blocks.
- Use one `selectBlock` for the canvas. `useCanvasUtils` had a copy without these checks. `BuilderCanvas` now passes the `useBlockSelection` version to `useCanvasUtils`.
- `scrollIntoView` no longer selects the block. It replaced the selection after a layer click that the new checks had ignored. `SearchBlock` now selects the result through `canvasStore.selectBlock`.

The layer highlight UI change moved to #846.


surajshetty3416:
@stravo1 not liking it as much... try removing border? and just try with different grays for selection instead of blue?

stravo1:
> @stravo1 not liking it as much... try removing border? and just try with different grays for selection instead of blue?

Split to https://github.com/frappe/builder/pull/846#issuecomment-5756490953

surajshetty3416 (approved):

PR #846, opened 2026-09-21
style: highlight selected layers and their nested layers

stravo1:
## Changes

- Give a selected layer a blue background. Use a darker blue in dark mode.
- Give the layers nested in a selected layer a gray background. The gray background is on the nested layer tree, so the rounded corners of a nested layer do not show the blue background.

Before this change, a single outline showed the selected block. The new backgrounds show more clearly which blocks are selected.

https://github.com/user-attachments/assets/df54d0d1-5456-4223-b266-a739b863fd76

Dark mode:

Screenshot 2026-09-17 at 9 34 30 PM

Split from #839.


stravo1:
Grays:
Screenshot 2026-09-18 at 11 05 37 AM

Blues:
Screenshot 2026-09-18 at 11 06 26 AM


https://github.com/user-attachments/assets/60e954b0-8a93-4769-95ee-ed629344289b



In my opinion we should stick to blue as we also show the actual block selection in the canvas with a blue border, this feels more consistent (also blue is the accent color of Builder).

